PR #6, opened 2021-09-06
added basic database testing + operations

Varun-Sethu:


Varun-Sethu:
will re-open in another PR

PR #12, opened 2021-09-07
Session/w21 94

fafnirZ:
updated branch and implemented some basic session management logic, with gorilla


Backend/go.mod
```diff
@@ -12,13 +12,17 @@ require (
...
+	github.com/gorilla/sessions v1.2.1
+=======
 	github.com/gotestyourself/gotestyourself v2.2.0+incompatible // indirect
 	github.com/jackc/pgtype v1.8.1
 	github.com/jackc/pgx/v4 v4.13.0
 	github.com/lib/pq v1.10.3
 	github.com/opencontainers/image-spec v1.0.1 // indirect
 	github.com/opencontainers/runc v1.0.2 // indirect
 	github.com/ory/dockertest v3.3.5+incompatible
+>>>>>>> prototype
```
Varun-Sethu, line 25: dont commit a merge conflict

Backend/go.sum
```diff
@@ -86,9 +86,16 @@ github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
...
+<<<<<<< HEAD
+github.com/gorilla/securecookie v1.1.1 h1:miw7JPhV+b/lAHSXz4qd/nN9jRiAFV5FwjeKyCS8BvQ=
+github.com/gorilla/securecookie v1.1.1/go.mod h1:ra0sb63/xPlUeL+yeDciTfxMRAA+MP+HVt/4epWDjd4=
+github.com/gorilla/sessions v1.2.1 h1:DHd3rPN5lE3Ts3D8rKkQ8x/0kqfeNmBAaiSi+o7FsgI=
+github.com/gorilla/sessions v1.2.1/go.mod h1:dk2InVEVJ0sfLlnXv9EAgkf6ecYs/i80K/zI+bUmuGM=
+=======
 github.com/gorilla/schema v1.2.0 h1:YufUaxZYCKGFuAq3c96BOhjgd5nmXiOY9NGzF247Tsc=
 github.com/gorilla/schema v1.2.0/go.mod h1:kgLaKoK1FELgZqMAVxx/5cbj0kT+57qxUrAlIO2eleU=
 github.com/gorilla/websocket v1.4.0/go.mod h1:E7qHFY5m1UJ88s3WnNqhKjPHQ0heANvMoAMk2YaljkQ=
+>>>>>>> prototype
```
Varun-Sethu, line 98: merge conflict

Backend/internal/auth/login.go
```diff
@@ -93,11 +98,14 @@ func LoginHandler(w http.ResponseWriter, r *http.Request) {
...
 		// else create a session if user's session isnt already created
-		// todo
-		//
-		// placeholder return success for now
-		// i know this is bad
-		throwErr(errors.New("success"), w)
+		_session.CreateSession(w, r, user.Email)
+
+		// will change to FRONTEND_URI soon
+		http.Redirect(w, r, "http://localhost:3000/dashboard", http.StatusMovedPermanently)
```
Varun-Sethu, line 104: switch statements require a break otherwise they will attempt to verify the next case

Backend/internal/auth/login.go
```diff
@@ -93,11 +98,14 @@ func LoginHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	case "DEFAULT":
```
Varun-Sethu, line 106: is this meant to be default?
```go
switch {
 case a:
    break
 default:
    break

}

```

Backend/internal/auth/login.go
```diff
@@ -132,7 +140,7 @@ func (u *User) checkPassword() error {
 	// i.e. even if the user puts ' or 1= '1
 	// the hashed value will not be an injection command
 
-	matches := CredentialsMatch(u.Email, hashedPassword)
+	matches := _db.CredentialsMatch(u.Email, hashedPassword)
 	if matches == 0 {
 		return errors.New("invalid credentials")
 	} else if matches > 1 { // if more than 1 result is returned
```
Varun-Sethu, line 146: one thing we could try is that maybe we could add a constraint to the schema? so that we never get into a situtation of duplicate entries? something like:
```sql
CONSTRAINT no_dupes UNIQUE (email, password)
```

Backend/internal/session/session.go
```diff
@@ -0,0 +1,116 @@
...
+	//required
+	gob.Register(User{})
+}
+
+// if session does not exist, create it
+// else return session
+func CreateSession(w http.ResponseWriter, r *http.Request, email string) {
+	session, err := store.Get(r, cookie_prefix)
+	if err != nil {
+		log.Println("an error has occurred getting session")
```
Varun-Sethu, line 53:
```go
log.Print("an error has occured getting session\n", err.Error())
```

Backend/internal/session/session.go
```diff
@@ -0,0 +1,116 @@
...
+
+}
+
+// returns user from session store
+// if there exists such a user
+func getUsers(s *sessions.Session) User {
+	val := s.Values["user"]
+	var user = User{}
+	user, ok := val.(User)
+	if !ok {
```
Varun-Sethu, line 112: looks good, another thing you could try (not required but its a syntax quirk is)
```go
if user, ok := val.(User); ok {
   return user
} else {
   return User{Authenticated: false}
}

```

Varun-Sethu (approved):
looks good, just fix up the switch statement and merge conflict

Backend/internal/auth/login.go
fafnirZ: oooo i see
ill look into that

Backend/internal/session/session.go
fafnirZ: roger

PR #13, opened 2021-09-08
W21 58 popupmenu

omeh-a:
This merge contains all the code for our modal "new file" window. Currently has all the infrastructure required to interpret a JSON list of templates and provide the user with options to select one of them and a title. Modal window is triggered by new file button in file area.

Note: currently no real infrastructure to display images for previews given that we will need to find a way to move these from the backend to the frontend for this.


cms_frontend/src/pages/Dashboard.tsx
```diff
@@ -19,6 +20,19 @@ const Directory = styled.h3`
 const Dashboard: React.FC = () => {
   const [dir, setDir] = useState("root" as FolderName);
 
+  // Modal state handler
+  const [open, setOpen] = React.useState(false);
+
+  // Modal opener
+  const handleOpen = () => {
```
fafnirZ, line 27: maybe name it more specificically
`handleModalOpen` might be good, since hanyuan's PR includes alot of functions in the dashboard

omeh-a: Will update. Thanks

PR #17, opened 2021-09-13
Logout/w21 95

fafnirZ:
Summary of this PR is as follows:

- whether session cookie is valid or not, will send a set-cookie header which will revoke any current session-token cookie stored in user brower, redirection will be done on client side
- this feature is tested to work and successfully removes the cookie from client's browser
- Added CORs support
- Changed docker-compose file to include automated running of infile so then changes to the infile (now renamed) will automatically update when `make` is run
- infile -> create_tables.sql (RENAME)


Backend/internal/auth/logout.go
```diff
@@ -0,0 +1,52 @@
...
+
+func LogoutHandler(w http.ResponseWriter, r *http.Request) {
+
+	switch r.Method {
+	case "POST":
+		authenticated, _ := _session.IsAuthenticated(w, r)
+		log.Print(authenticated)
+		if authenticated {
+			// CORS headers
+			w.Header().Set("Access-Control-Allow-Origin", "http://localhost:3001")
```
Varun-Sethu, line 31: super small nitpick but maybe make this a constant just so that if we deploy this then we dont need to go through all change all instances of localhost

Backend/internal/auth/logout.go
```diff
@@ -0,0 +1,52 @@
...
+			// is storing by returning a
+			// Set-Cookie: session-token="" header
+			w.Header().Set("Access-Control-Allow-Origin", "http://localhost:3001")
+			w.Header().Set("Access-Control-Allow-Credentials", "true")
+			_session.RemoveSession(w, r)
+			_httpUtil.ThrowRequestError(w, http.StatusUnauthorized, "unauthorized")
+			break
+		}
+
+	case "DEFAULT":
```
Varun-Sethu, line 46: is this meant to be "DEFAULT" or default: ?
```go
switch {
 case A:
 default:
}
```

Varun-Sethu (approved):
everything looks good, maybe just check the switch statement though

Backend/internal/auth/logout.go
fafnirZ: ooooo oops HAHAHHA

Backend/internal/auth/logout.go
fafnirZ: yeah truee, I was thinking maybe we could use an environment variable maybe??

Varun-Sethu: yah env variable sounds good coz we can also pass it in with docker

PR #18, opened 2021-09-13
Implement new folder functionality

hanyuone:
Fixes W21-98, also implements basic renaming functionality for files/folders.

Note that in this PR renaming isn't recursive, i.e. if a folder is renamed the child files aren't renamed - however, this is a POC and this functionality should be done in the backend anyway.


hanyuone:
@omeh-a there's a bunch of merge conflicts with your modal code for new files, not sure which one is best to keep

cms_frontend/src/components/FileRenderer/FileRenderer.tsx
```diff
@@ -42,34 +45,37 @@ const sortFiles = (files: FileFormat[]) => {
 // Typescript Declaration
 // uses the interface defined above
 // imports the icons from material UI
-const FileRenderer: React.FC<RenderProps> = ({ files, onFileClick, onFolderClick, onNewFile }) => {
+const FileRenderer: React.FC<RenderProps> = (props) => {
+  const { files, onFileClick, onFolderClick, onRename, onNewFile } = props;
```
fafnirZ, line 49: lets keep the style to be consistenly ({props}) instead of destructuring it afterwards.

cms_frontend/src/pages/Dashboard.tsx
```diff
@@ -41,45 +38,146 @@ const Dashboard: React.FC = () => {
...
   // Moves our current directory up (analogous to `cd ..`)
   const toParent = () => {
-    const parent = getParent() as FolderName;
+    const parent = getParent();
     setDir(parent);
+    setFolder(Files.get(parent) as FileFormat[]);
   }
 
   // Checks if our current directory has a parent directory
   const hasParent = () => {
```
fafnirZ, line 47: A bit redundant imo, since all folders has a parent except for root dir, so maybe instead of using a for loop just check that the CWD is root dir?

cms_frontend/src/pages/Dashboard.tsx
```diff
@@ -41,45 +38,146 @@ const Dashboard: React.FC = () => {
...
     const parent = getParent();
+    let found = false;
+
+    Files.forEach((_, key) => {
```
fafnirZ, line 51: A bit redundant imo, since all folders has a parent except for root dir, so maybe instead of using a for loop just check that the CWD is root dir?

cms_frontend/src/pages/Dashboard.tsx
```diff
@@ -41,45 +38,146 @@ const Dashboard: React.FC = () => {
...
+    Files.set(dir, updated);
+  }
+
+  // Finds the next available folder name
+  const newFolderName = () => {
+    let index = 0;
+    let folder_name = "New Folder";
+
+    while (containsFilename(folder_name, "folder")) {
+      index++;
```
fafnirZ, line 84: I'm quite confused as to what this does, and whether or not it will end up in an infinite loop

hanyuone: This generates a name for a new folder ("New Folder", "New Folder (1)", "New Folder (2)") and uses the first available name, shouldn't infinite loop since a directory only has a finite number of files so we'll eventually find a suitable name

cms_frontend/src/pages/Dashboard.tsx
hanyuone: Yep, will change

cms_frontend/src/pages/Dashboard.tsx
fafnirZ: right Okay i didnt see where folder_name was coming from, all goods looks good then!!

omeh-a (approved):
Looks good - Jacky's feedback addresses everything i would have flagged :)

PR #20, opened 2021-09-17
W21 75 filesystemchildren

omeh-a:
This PR implements a function which pulls all entity children as a list of IDs. This function uses the children row in the db, so it is safe to remove the parent row in the future without changes to this property. Unit tests are included and there is little else to report for this :)

### DIFF
--- a/Backend/filesystem/service.go
+++ b/Backend/filesystem/service.go
@@ -4,6 +4,7 @@ import (
 	"DiffSync/database"
 	"context"
 	"errors"
+	"strconv"
 
 	"github.com/jackc/pgtype"
 )
@@ -97,8 +98,8 @@ func renameEntity(pool database.Pool, entityID int, newName string) error {
 	return err
 }
 
-// getFilesystemInfo returns information regarding a specific file system entity
-func getEntityChildren(pool database.Pool, docID int) ([]string, error) {
+// getFilesystemChildren returns the list of children for a file system entity
+func getEntityChildren(pool database.Pool, docID int) ([]EntityInfo, error) {
 	children := pgtype.Hstore{}
 
 	err := pool.GetConn().QueryRow(context.Background(), "SELECT children FROM filesystem WHERE entityid = $1",
@@ -107,10 +108,12 @@ func getEntityChildren(pool database.Pool, docID int) ([]string, error) {
 		return nil, errors.New("failed to read from database")
 	}
 
-	list := []string{}
+	list := []EntityInfo{}
 
 	for k := range children.Map {
-		list = append(list, k)
+		id, _ := strconv.Atoi(k)
+		info, _ := getFilesystemInfo(pool, id)
+		list = append(list, info)
 	}
 
 	return list, nil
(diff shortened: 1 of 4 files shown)


Backend/filesystem/service.go
```diff
@@ -97,20 +97,21 @@ func renameEntity(pool database.Pool, entityID int, newName string) error {
...
+	err := pool.GetConn().QueryRow(context.Background(), "SELECT children FROM filesystem WHERE entityid = $1",
+		docID).Scan(&children)
+	if err != nil {
+		return nil, errors.New("failed to read from database")
+	}
+
+	list := []string{}
+
+	for k := range children.Map {
+		list = append(list, k)
```
Varun-Sethu, line 113: This isn't what `getEntityChildren` should be doing, the functionality is close though. When the frontend calls the getEntityChildren endpoint ideally we should also return some basic information regarding each child (not just the ID), that way they don't have to repeatedly flood the backend with more requests when trying to capture information about each child. Ideally the response should look something like:
```json
{
   "children": [
      {
          "entityID": 1234,
          "name": "cool document",
          "isDocument": true
      },
      {
          "entityID": 1234,
          "name": "cool directory",
          "isDocument": false
      },
   ]
}
```

Backend/filesystem/service_test.go
```diff
@@ -140,3 +140,46 @@ func TestEntityRename(t *testing.T) {
...
+	d2_kids, _ := getEntityChildren(pool, dir2)
+	d3_kids, _ := getEntityChildren(pool, dir3)
+	d4_kids, _ := getEntityChildren(pool, dir4)
+	de_kids, _ := getEntityChildren(pool, emptyDir)
+
+	assert.True(len(d1_kids) == 3)
+	assert.True(len(d2_kids) == 1)
+	assert.True(len(d3_kids) == 4)
+	assert.True(len(d4_kids) == 9)
+	assert.True(len(de_kids) == 0)
```
Varun-Sethu, line 183: tests look good, maybe add a test to verify that the actual contents is also wat we expect (dont have to do it for all; just one)

Backend/filesystem/service.go
Varun-Sethu: so something roughly like:
```go
children := []EntityInfo{}

for k := range children.Map {
   info, _ := getFilesystemInfo(pool, k)
   children = append(children, info)
}
return children
```

Varun-Sethu (changes requested):
Just need to append entity information to the array as opposed to just EntityIDs

Backend/filesystem/service_test.go
omeh-a: oooh my bad,  will update to match the api docs

Backend/filesystem/service.go
omeh-a: oooh my bad, will update to match the api docs

Backend/filesystem/service_test.go
omeh-a: Is it totally necessary to do so if the contents are just a return from getFilesystemInfo? The only new data this func returns is the actual children themselves

Backend/filesystem/service.go
omeh-a: I've changed it to return entityinfo now, although that also means including an array of the children of the children since it's using the output from getFilesystemInfo

PR #22, opened 2021-09-21
updated groups schema to include permissions as well as creating a pe…

fafnirZ:
…rmissions helper function in auth package.

wont be creating any more permission related functions yet until we implement the edit page route. where we actually need to check whether or not the user can access the file.


Backend/internal/auth/auth_DB.go
```diff
@@ -0,0 +1,53 @@
...
+func init() {
+	var err error
+	httpDbPool, err = database.NewPool(database.Config{
+		HostAndPort: "db:5432",
+		User:        "postgres",
+		Password:    "postgres",
+		Database:    "test_db",
+	})
+
+	if err != nil {
```
omeh-a, line 30: Where do these print statements end up? Should we maybe be implementing exceptions instead?

Postgres/create_tables.sql
```diff
@@ -1,33 +1,48 @@
 CREATE EXTENSION hstore;
 SET timezone = 'Australia/Sydney';
 
+CREATE TABLE groups (
+  UID   SERIAL PRIMARY KEY,
+  Name  VARCHAR(50) NOT NULL,
+  Permission int UNIQUE NOT NULL
```
omeh-a, line 7: May be worth adding a procedure or something which handles the permission level - this will be pretty annoying to manage if it persists to the end product.

Backend/internal/auth/auth_DB.go
fafnirZ: the print statements end up in the console, but ye I could look into that

Postgres/create_tables.sql
fafnirZ: yep sure

Backend/config/config.go
```diff
@@ -0,0 +1,10 @@
+package config
+
+import (
+	"os"
+)
+
+func GetFrontendURI() string {
+	env := os.Getenv("FRONTEND_URI")
+	return env
```
Varun-Sethu, line 9: can also just 
```go
return os.Getenv("FRONTEND_URI")
```

Backend/internal/auth/auth_DB.go
```diff
@@ -0,0 +1,53 @@
...
+	"log"
+
+	_ "github.com/lib/pq"
+)
+
+var httpDbPool database.Pool
+
+func init() {
+	var err error
+	httpDbPool, err = database.NewPool(database.Config{
```
Varun-Sethu, line 23: isnt there meant to be environment variables for the database too?

Backend/internal/auth/auth_DB.go
Varun-Sethu: it might be beneficial to have a config validator that runs prior to the backend accepting any requests, the validator can just checks that the provided DB config details actually results in a working connection

Postgres/create_tables.sql
Varun-Sethu: also, enums will make it easier to maintain: https://www.postgresql.org/docs/9.1/datatype-enum.html

Varun-Sethu: so 
```plpgsql
CREATE TYPE permissions_enum AS ENUM ('read', 'write', 'delete');
# ... 
# ...
Permission permissions_enum UNIQUE NOT NULL
```
or something

Cms_frontend/src/App.tsx
```diff
@@ -5,6 +5,7 @@ import './css/styles.css'
 
 
 const App: React.FC = () => {
+
```
Varun-Sethu, line 8: ebic

Backend/internal/auth/login.go
```diff
@@ -91,11 +67,11 @@ func LoginHandler(w http.ResponseWriter, r *http.Request) {
 		// will change to FRONTEND_URI soon
 		//_httpUtil.SendResponse(w, "success")
 
-		http.Redirect(w, r, "http://localhost:3000/dashboard", http.StatusMovedPermanently)
+		http.Redirect(w, r, FRONTEND_URI, http.StatusMovedPermanently)
```
Varun-Sethu, line 70: should this and line 74 redirect to the same location? or should it be something like:
```go
FRONTEND_URL + "/dashboard"
```
and
```go
FRONTEND_URL + "/login"
```

Varun-Sethu (changes requested):
I was under the impression that environment variables would also be passing in db config?

PR #26, opened 2021-09-29
Created Editor UI

psoo1:
Created Editor UI according to Figma design.


omeh-a (approved):

omeh-a:
I added on a sidebar element and also reorganised the component files just to make things a bit neater. I figured that it's probably easiest for everybody else if we have the sidebar just *there* so there doesn't need to be another branch for it (or lots of merge conflicts later).

Cms_frontend/src/components/Editor/EditorSidebar.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import styled from 'styled-components';
+
+
+const Container = styled.div`
+  background: #eeeeee;
+  width: 15%;
```
fafnirZ, line 7: i think viewport width most of the time are better because i think 15% takes 15% of the parent's width

Cms_frontend/src/components/Editor/EditorSidebar.tsx
```diff
@@ -0,0 +1,20 @@
...
+  overflow-y: scroll;
```
fafnirZ, line 8: i think viewport width is better because % takes % of parent's width and sometimes the parent's width isnt 100% of screen size, overflow-y: i guess we can add a scroll but, imo we wont have enough items and even if we do have that many items im not sure if a scrollbar is the best UI design for it

Cms_frontend/src/pages/Editor.tsx
```diff
@@ -10,11 +11,16 @@ const Container = styled.div`
 `
 const Editor: React.FC = () => {
     return (
-        <div className="Editor">
+        <div style={{ height: '100%' }}>
```
fafnirZ, line 14: another way that might be helpful is instead of using inline styles, we could import `{ Box }` from material ui and then just write `<Box height="100%">` , just something i personally like to do nowadays instead of using inline style tags

Cms_frontend/src/components/Editor/EditorSidebar.tsx
omeh-a: Just included the scrollbar because that's what we had on our figma

PR #27, opened 2021-10-01
Upgrading how database tests are performed

Varun-Sethu:


Varun-Sethu:
# Upgrading Database Tests
This is quite a big PR and I'm really sorry about that, in general the following changes are being introduced:
 - Improved database testing
 - Moving config schema out to a file and reading from there
 - Adding Github Actions
 - Quick patch to the `filesystem/info/root` endpoint

## Improved Database Testing
As discussed on the Discord server database queries and tests are now performed via a `DatabaseContext` interface, HTTP methods have a context backed by a live DB while test methods have a context backed by a special testing database. Additionally, anything run under `go test ./...`  will not be able to touch the live database. To perform a test on a database bound method we now instantiate a new TestingContext and pass our test as a lambda method to the `RunTest` function like so:
```go
func TestSomeCoolStuff(t *testing.T) {
   ctx := database.NewTestingContext()
   ctx.RunTest(func () {
       testDbBoundMethod(ctx)
   })
}
```
after a test is performed the changes are rolled back and isolated from other tests. If you want to perform a test that might fail you must additionally wrap test around a `WillFail` method like so:
```go
func TestSomeCoolStuff(t *testing.T) {
   ctx := database.NewTestingContext()
   ctx.RunTest(func () {
       assert.True(ctx.WillFail(func() error { 
            return badTest(ctx)
       }))
   })
}
```

Additionally, all tests are now performed against a testing database (specified in the `docker-compose.yml`), if someone attempts to run tests locally however then a new docker container containing a test DB will be spun up (if the test DB isn't active)

## Moving Config Schema
Before the schema for the testing database was stored as a constant, now its just read directly from `create_tables.sql`

## Adding Github Actions
We now have automated testing 🥳 

## Endpoint Patch
(Yes).

Backend/database/liveContext.go
```diff
@@ -0,0 +1,54 @@
+package database
```
omeh-a, line 1: Might be a good idea to include some header comments in these files, because i think these ones are especially likely to prompt next  year's people to ask for help

Backend/database/tests/database_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// I'm so sorry about this
+	assert.NotPanics(func() {
+		testContext.RunTest(func() {
+			testContext.Exec("DROP TABLE IF EXISTS person", []interface{}{})
+		})
+	})
+}
+
+func TestUpdatesAreIsolated(t *testing.T) {
```
omeh-a, line 46: Add some comments describing when these functions should be called and a super brief description

omeh-a (changes requested):
Looks good, although you need to take a pass at this to add comments to one or two of these files and at least copypasta a header comment onto these files with your contact details, given that you likely are going to remain the authority on these files for quite a bit of time to come and people almost certainly will need to ask you about it sometime

omeh-a (approved):

PR #30, opened 2021-11-19
Basic integration of DraftJS

hanyuone:
Added a barebones version of the DraftJS editor into the `/editor` page, and a "bold" button as a proof-of-concept. The style for the DraftJS editor itself might be a bit hacky, but it works for now (the previous version would extend beyond the width of the page).


Cms_frontend/src/components/Editor/EditorSidebar.tsx
```diff
@@ -1,20 +1,40 @@
...
 
 const Container = styled.div`
   background: #eeeeee;
   width: 15vw;
   overflow-y: scroll;
 `
 
+interface EditorProps {
+  editorState: EditorState,
+  setEditorState: React.Dispatch<React.SetStateAction<EditorState>>
```
fafnirZ, line 14: damn interesting, I didnt know that was the types for it

fafnirZ:
Overall looks good, can you chuck a quick screenshot for what it looks like here so we can have a look at how it looks visually?

Cms_frontend/package.json
```diff
@@ -8,10 +8,12 @@
     "@testing-library/jest-dom": "^5.14.1",
```
omeh-a, line 8: Looks good overall. There are a few ESLint flags but we can be lenient about it since this is just a placeholder anyway.

omeh-a (approved):

PR #31, opened 2021-12-09
Backend Editor Refactor + Moving Code to main

Varun-Sethu:
I apologise for the size of this PR although do note that a lot of it is just moving code around, majority of the **important** changes are within the editor package 😄 .


Varun-Sethu:
IMO there are some current issues with the current design for the editor, those are as follows:
**Some notable issues**

- The event loop is extremely confusing and calling certain functions results in a deadlock, this is not scalable at all and we need to think up a better solution (potentially only marking certain methods as usable in an event loop?)
- Is the singleton pattern a good idea?
- Should the document manager expose a document closure method? Or just start and spin documents?

Maybe they should be resolved later in another Jira task? The code is good enough for main (for now)

Varun-Sethu:
I think one good example of that deadlock is when calling `doc.close()` within the event loop, since our channels are blocking if we call the function within the event loop it will result in the current go-routine being moved out of the scheduling queue as it waits for another go-routine to collect the data it just pushed; the issue however is that the go routine that would collect the data in the channel is the routine that was just put to sleep, resulting in the entire document dying.

Perhaps multiple go-routines for different event types is a better idea? But that seems odd, or instead we just stop people from calling event-loop functions from within the event-loop but that does not scale well, maybe the event loop channels become non-blocking instead but concurrency challenges may become annoying 😭

Varun-Sethu:
If you give the editor architecture document a quick read too you will see that documents are responsible for terminating themselves when their extension count reaches 0, perhaps this should be done by the manager (seems a lot more natural); this change isnt hard to add but tbh I feel like it should be a separate jira task. 

The change just involves refactoring extension initialisation a bit

.vscode/dryrun.log
```diff
@@ -0,0 +1,7 @@
+make --dry-run --always-make --keep-going --print-directory
+make: Entering directory '/home/varun/dev/go/src/cms.csesoc.unsw.edu.au'
```
fafnirZ, line 2: are the .vscode files necessary? it contains a path to your home directory

Backend/database/dbFetcher.go
```diff
@@ -0,0 +1,149 @@
+// TITLE: DB Fetcher
+// Created by (Varun: Varun-Sethu) (09/21)
+// Last modified by (Varun: Varun-Sethu) (1/10/21)
+// # # #
+/*
+	This file is involved with fetching/spinning up databases to host unit tests
+	on
```
fafnirZ, line 7: comment is a little incomplete just a heads up

.vscode/dryrun.log
Varun-Sethu: ohh damm didnt notice htat

Backend/editor/service/autosave_extension.go
```diff
@@ -0,0 +1,61 @@
...
+// Spin in a loop listenining for updates on our
+// websocket connection
+func (c *AutosaveHead) Spin() {
+	for {
+		switch {
+		case <-c.stopSpinning:
+			return
+		default:
+			// TODO: write to file system every n seconds
+			// @Jacky
```
fafnirZ, line 53: roger that

Backend/editor/service/broker.go
```diff
@@ -0,0 +1,68 @@
...
+
+	for _, extName := range requiredExtensions {
+		b.manager.LoadExtension(documentID, extensionFactory(
+			extName, conn,
+		))
+	}
+	return nil
+}
+
+// small factory for generating extensions based on a name :)
```
fafnirZ, line 55: damn factory pattern AND singleton pattern HAHAHAHA

Backend/filesystem/endpoints.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+type ValidRenameRequest struct {
+	EntityID int    `schema:"EntityID,required"`
+	NewName  string `schema:"NewName,required"`
+}
+
+// Handler for renaming filesystem entities
+func RenameFilesystemEntity(w http.ResponseWriter, r *http.Request) {
```
fafnirZ, line 132: the casing here is a bit off houldnt it be RenameFileSystemEntity??

fafnirZ (commented):
lgtm :D

fafnirZ (approved):

Backend/filesystem/endpoints.go
Varun-Sethu: Filesystem seems to be consistent casing

fafnirZ: Yeah I realised afterwards oops

PR #32, opened 2021-12-25
String algorithms + WASM stubs

Varun-Sethu:
Algorithms for:
 - Suffix/Prefix
 - Myer's string diff generation

There will also be another PR to extend these to JSON objects factoring in their unique syntax


Backend/algorithms/strings.go
```diff
@@ -0,0 +1,211 @@
...
+// CommonPrefix returns the common prefix between two
+// sentences represented as an array of words
+func CommonPrefix(a, b []string) int {
+	bound := min(len(a), len(b))
+
+	// allow compiler to eliminate
+	// bounds check
+	_ = a[bound-1]
+	_ = b[bound-1]
+
```
fafnirZ, line 53: can you explain what you mean by eliminating bounds checks?

fafnirZ (approved):
lgtm

PR #39, opened 2022-02-28
fixing eslint bug and broken BE

fafnirZ:


Varun-Sethu:
Looks good, there was another change in your last PR regarding a change in the output of one of the endpoints, the test corresponding to one of the endpoints is still failing so just fix that before merging

Varun-Sethu (approved):
As said in comment, looks good; just update test

PR #42, opened 2022-03-04
Removing Children HStore from Schema

Varun-Sethu:
Finally :D


fafnirZ:
LGTM, but can you give me a little more context as to why it was stored as a hstore (from my little understanding, hstore is kind of like a dictionary in sql?) why was that bad?

PR #43, opened 2022-03-04
[CMS-226] Adding more storybook templates, Adding more CI/CD checks for frontend

fafnirZ:
## Why it was required?
- Frontend needed a way to visually render and test resuable components completely isolated from the integreation of the component with the rest of the app, that is why we are using storybook
- Frontend also required a way to prevent introducing breaking changes to main branch, and running a pre-push type check and eslint check will help prevent small things from breaking main

## Changes
- Added storybook and created a template for it
- Added more pre push CI/CD steps for FE

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,4 +29,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: type check
-        run: cd Frontend && npm run types
+        run: cd Frontend && npm install && npm run types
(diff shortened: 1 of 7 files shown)


Varun-Sethu:
Why are the frontend tests failing?

fafnirZ:
Its saying the import path for an import is incorrect, will try and fix it soon

.github/workflows/ci.yml
```diff
@@ -24,3 +24,9 @@ jobs:
       - name: Golang Tests
         run: go test -v ./...
         working-directory: ./Backend
+  frontend_testing:
```
Varun-Sethu, line 27: if we're calling this frontend_testing perhaps we should call the other one backend_testing

Varun-Sethu (approved):

PR #66, opened 2022-03-20
[EASY][BUG] Filename rename doesnt actually change the name

james-teng-0:

- after creating a new page and trying to rename the page for the first time, the rename doesn't work (after entering the rename, it goes back to the initial name)
- this was because the page didn't have a valid ID intially.

## **Changes**
- Changed api for newFile to return the ID instead of JSON


fafnirZ (approved):
Lgtm!! 😄

PR #75, opened 2022-03-26
[CMS-140] Adding react testing library to the repository

fafnirZ:
## Why was it required?
Our code repository required a way to test our frontend, react testing library is a great library used to provide integration tests, which tests whether components work how they are supposed to work when interacting with one another

## Changes
- added setup code which allows us to call tests with custom queries
- added a few example tests that pass
- added to tsconfig
- added @testing-library/react to dev dependancies


Frontend/src/cse-testing-lib/custom-queries.ts
```diff
@@ -0,0 +1,32 @@
+import {queryHelpers, buildQueries} from '@testing-library/react'
+
+// The queryAllByAttribute is a shortcut for attribute-based matchers
+// You can also use document.querySelector or a combination of existing
+// testing library utilities to find matching nodes for your query
+
+const queryAllByDataAnchor = (...args: any[]) => {
+  // @ts-ignore
+  return queryHelpers.queryAllByAttribute('data-anchor', ...args)
+}
```
zijizhu, line 10: Hi, eslint complains about this function: eslint[@typescript-eslint/explicit-module-boundary-types] and eslint[@typescript-eslint/ban-ts-comment]
When I hover my mouse over the `buildQueries` function below I see react-testing-library type this function as `GetAllBy<[dataAnchorValue: any]>`. So I tried this and it gets rid of all the eslint errors:
```
import {queryHelpers, buildQueries, GetAllBy} from '@testing-library/react'

const queryAllByDataAnchor : GetAllBy<[dataAnchorValue: any]> = (...args) => {
  return queryHelpers.queryAllByAttribute('data-anchor', ...args)
}
```

zijizhu (approved):
Looks good to me!

Frontend/src/cse-testing-lib/custom-queries.ts
fafnirZ: great pickup!! :D will update with this instead of the eslint ignore

PR #119, opened 2022-04-25
[CMS-245] implemented elements and added to the homepage

leannachan:
The homepage look like this:
image

What I did:
- exported SVG of csesoc logo and added to the homepage
- styled the text such as italic, shadow effect, font family
- added the button


fafnirZ (approved):
lgtm

PR #132, opened 2022-05-27
[CMS-175] Adding component factory to Next.js

zijizhu:
changes:

- Implemented component factory (`Next/src/components/Blog.tsx`) which parses the proposed blog document data structure and displays the concent. The data structure is defined in `Next/src/types.ts`.
- Added a /blog route which shows an example blog
- Style fixes and cleanups for Next.js index page


Next/src/components/Blog.tsx
```diff
@@ -0,0 +1,66 @@
...
+                {text}
+              </Link>
+            </StyledLink>
+          ) : (
+            <>{text}</>
+          )}
+        </Text>
+      ))}
+    </ParagraphBlock>
+  );
```
fafnirZ, line 53: potential use of switch case instead to improve code clarity. the ternery if else rendering is a little confusing. Otherwise commenting should be sufficient

Next/src/pages/blog/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+        bold: true,
+        italic: true,
+      },
+      {
+        text: ", and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.",
+        bold: true,
+      },
+    ],
+  },
+];
```
fafnirZ, line 77: move this into a blog data mock file instead of placing it into the `index.tsx` file

Next/src/components/Blog.tsx
```diff
@@ -0,0 +1,66 @@
...
+`;
+
+const ParagraphBlock = styled.p<ParagraphStyle>`
+  text-align: ${(props) => props.align ?? "left"};
+`;
+
+const BlogContainer = styled.div`
+  max-width: 600px;
+  font-size: 1.25rem;
+`;
```
fafnirZ, line 31: since there is greater than 4 different types of styles, would make sense to chuck it in a `-styled.tsx` file

Next/src/types.ts
```diff
@@ -0,0 +1,29 @@
...
+interface Text extends TextStyle {
+  text: string;
+  link?: string;
+}
+
+export interface TextStyle {
+  bold?: boolean;
+  italic?: boolean;
+  underline?: boolean;
+}
```
fafnirZ, line 29: since this is the types for the component factory only and not for the entire NextJS application wouldn't it make sense to be inside a `components/Blog` folder instead?

fafnirZ (changes requested):
a couple of comments to address but otherwise lgtm

Next/src/pages/blog/index.tsx
zijizhu: Yeah sure I totally agree with this

Next/src/components/Blog.tsx
zijizhu: > potential use of switch case instead to improve code clarity. the ternery if else rendering is a little confusing. Otherwise commenting should be sufficient

Since there are only two cases for the link attribute, a ternery expression is more appropriate here IMO. However I do think swith case is very handy for other stuff like paragraph alignment (in the future). I will add some comments to this component to make it a bit clearer.

fafnirZ: aight sounds good

Next/src/components/blog/Styled.tsx
```diff
@@ -0,0 +1,32 @@
...
+const ParagraphBlock = styled.p<ParagraphStyle>`
+  text-align: ${(props) => props.align ?? "left"};
+`;
+
+const BlogContainer = styled.div`
+  max-width: 600px;
+  font-size: 1.25rem;
+`;
+
+export { Text, StyledLink, ImagePlaceholder, ParagraphBlock, BlogContainer };
```
fafnirZ, line 32: A bit nit picky about naming conventions of this file, maybe `Blog-styled.tsx` makes more sense to highlight that this is the styles for the `Blog.tsx` component

fafnirZ (approved):

PR #135, opened 2022-05-28
[CMS-271] Create swirl curvecomponents in figma and import into homepage

fafnirZ:
## why this was necessary?
To complete design

## Changes
- added swirls as customisable svg
- added curve containers

## Screenshots
image

image

image


james-teng-0 (commented):
use p instead of body tag for text elements

PR #136, opened 2022-05-28
[CMS-221] edit after clicking onto selected file navigates to editor with correct fileid

fafnirZ:
## Why changes were necessary
It is necessary to link up the edit button with the associated file editor

## Changes
- added parameters to path 
- added logic for selecting file and associating that with the edit button


james-teng-0:
looks good - (I added a change to the test file to pass the checks)

james-teng-0:
looks good - (I added a routes tag to the siderbar test file to pass the checks)

PR #139, opened 2022-05-29
[CMS-183]: Integrate Redux to Editor

kylee-software:
Changes:
- Integrated Redux to the Editor so that changes to content blocks are stored and updated in Redux
- When a content block is added, Redux will create an initial state for it
- Data of each content block is being stored as `BlockInfo` which has `id` and `data` as its properties. `data` is in JSON format
- Enable key and mouse events to bolden, italicize, and underline texts


fafnirZ:
Hi kylee, your PR seems to revert some of the more updated UI changes 

an image from [CMS-183]
image

an image from [main]
image

Please work with @zijizhu  to figure out what has happened and re-implement the UI that is in main right now.

PR #141, opened 2022-05-31
[CMS-256]: Integrate Redux to Editor

kylee-software:
Changes:

- Integrated Redux to the Editor so that changes to content blocks are stored and updated in Redux
- When a content block is added, Redux will create an initial state for it
- Data of each content block is being stored as `BlockInfo` which has `id` and `data` as its properties. `data` is in JSON format


zijizhu:
Looks good to me! 👍 Sinced the type is fixed, I don't have anything to add.

zijizhu (approved):

fafnirZ (approved):

PR #142, opened 2022-06-03
[CMS-258] Added "events" page section

lauraw0:

- The CSESoc website design includes a "events" page that provides an outline of CSESoc events and a cover photo of a recent/upcoming event.

## **Changes**

- Created an "events" page component in `Events.tsx` 
- Included an example cover photo (CSESocEventsCP.png)

## **Screenshots**
- The purple background will be later removed and is only temporary
image


Frontend/src/cse-ui-kit/events_container/EventsContainer.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
fafnirZ, line 1: being a little pedantic about the naming of this file, not sure if EventsContainer is a good name for this component if we want it to be a reusable component. Maybe something like `TransparentGlassContainer` etc. think of a good name, something that better describes what it looks like instead of EventsContainere

fafnirZ (approved):
lgtm, address changes and the FAILING tests then merge whenever
image

PR #146, opened 2022-06-04
Image Upload Interface

brianzhang2101:
Added CRUD helper functions to write to the "document_data" Docker volume


Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
 )
 
+const volume_path = "/var/lib/documents/data"
+
 // Implements IRepositoryInterface
 type FilesystemRepository struct {
 	embeddedContext
 }
 
+type DockerFileystemRepository struct {
```
angary, line 22:
```suggestion
type DockerFilesystemRepository struct {
```

angary: Also adjust spelling for other occurrences within this file

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+
+// Delete file from volume
+func (c *DockerFileystemRepository) DeleteFromVolume(filename string) error {
+	filepath := filepath.Join(volume_path, filename)
+	file, err := os.Open(filepath)
+	if err != nil {
+		return fmt.Errorf("File doesn't exist")
+	}
+	file.Close()
+	os.Remove(filepath)
```
angary, line 83:
```suggestion
	if err = os.Remove(filename); err != nil {
		return fmt.Errorf("Couldn't remove the source file")
	}
```
`os.Remove` returns an error - make sure to include this error check to match code above

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+
+// Create instance of DockerFileSystemRepository struct
+func NewDockerFilesystemRespository() (c *DockerFileystemRepository, err error) {
+	c = new(DockerFileystemRepository)
+
+	c.cli, err = client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return nil, err
+	}
+	return c, nil
```
angary, line 34:
```suggestion
    return client.NewClientWithOpts(client.FromEnv)
```
Is there a purpose for this if statement?

- If err is not nil, then we convert the `*DockerFilesystemRepository` to `nil` in the return value and return error as it already is (unlike other cases where you seem to change the error message)
  - This is not required as if the error handling is implemented correctly when this function is called, we wouldn't do anything with the `*DockerFilesystemRepository` when we catch the error
- If the err is nil, then we return `nil` instead of the error and return c normally
  - This doesn't do anything since the error is already nil

Hence you should be able to return the values as is.

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+	moved, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
+	defer moved.Close()
+	if err != nil {
+		return fmt.Errorf("Couldn't read/create the destination file")
+	}
+	// Copy source to destination
+	_, err = io.Copy(moved, src)
+	if err != nil {
+		return fmt.Errorf("File couldn't be copied to destination")
+	}
```
angary, line 55:
```suggestion
	if _, err = io.Copy(moved, src); err != nil {
		return fmt.Errorf("File couldn't be copied to destination")
	}
```

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+	src.Close()
+	// Delete source file
+	err = os.Remove(filename)
+	if err != nil {
+		return fmt.Errorf("Couldn't remove the source file")
+	}
```
angary, line 61:
```suggestion
	if err = os.Remove(filename); err != nil {
		return fmt.Errorf("Couldn't remove the source file")
	}
```

angary (changes requested):

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+}
+
+// Delete file from volume
+func (c *DockerFileystemRepository) DeleteFromVolume(filename string) error {
+	filepath := filepath.Join(volume_path, filename)
+	file, err := os.Open(filepath)
+	if err != nil {
+		return fmt.Errorf("File doesn't exist")
+	}
+	file.Close()
```
angary, line 82: `file.Close()` returns an error - not sure if this needs to be checked / handled though

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
 
+const volume_path = "/var/lib/documents/data"
+
 // Implements IRepositoryInterface
 type FilesystemRepository struct {
 	embeddedContext
 }
 
+type DockerFileystemRepository struct {
+	cli *client.Client
```
Varun-Sethu, line 23: perhaps this should be changed to:
```go
dockerCli *client.Client
```
as its not immediately clear what this is for

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+}
+
+// Create instance of DockerFileSystemRepository struct
+func NewDockerFilesystemRespository() (c *DockerFileystemRepository, err error) {
+	c = new(DockerFileystemRepository)
```
Varun-Sethu, line 28: nitpick: DockerFilesystemRepository is a struct type so it doesn't have to be initialized with ```new()```, instead you can just allocate it locally and return a pointer to it, during the escape analysis phase of compilation the compiler will automatically mark the variable as allocated on the heap for you:

```go
func myFunc() *DockerFilesystemRepository {
    myStruct := DockerFilesystemRepository {}
    return &myStruct
}

```

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
 type FilesystemRepository struct {
 	embeddedContext
 }
 
+type DockerFileystemRepository struct {
+	cli *client.Client
+}
+
+// Create instance of DockerFileSystemRepository struct
+func NewDockerFilesystemRespository() (c *DockerFileystemRepository, err error) {
```
Varun-Sethu, line 27: do the return values need to be named?

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+	defer moved.Close()
+	if err != nil {
+		return fmt.Errorf("Couldn't read/create the destination file")
+	}
+	// Copy source to destination
+	_, err = io.Copy(moved, src)
+	if err != nil {
+		return fmt.Errorf("File couldn't be copied to destination")
+	}
+	src.Close()
```
Varun-Sethu, line 56: you can defer this immediately after the open:
```go
src := os.Open(...)
defer src.Close()
```

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+	// Delete source file
+	err = os.Remove(filename)
+	if err != nil {
+		return fmt.Errorf("Couldn't remove the source file")
```
Varun-Sethu, line 60: nitpick: does this need to be formatted? Can do ```errors.New("Couldn't remove the source file")```

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,87 @@ package repositories
...
+	}
+	return nil
+}
+
+// Get file from volume. Returns a valid file pointer
+func (c *DockerFileystemRepository) GetFromVolume(filename string) (*os.File, error) {
+	// Concatenate volume path with file name
+	file, err := os.Open(filepath.Join(volume_path, filename))
```
Varun-Sethu, line 68: Similar thing to Gary's comments in ur previous function:
```go
if file, err := os.Open(filepath.Join(volume_path, filename)); err != nil {
   return nil, err
} else {
  return file, nil
}

```

Backend/database/repositories/filesystem.go
Varun-Sethu, line 82: Please defer this instead with: 
```go
defer file.Close()
```

Varun-Sethu: You can also ignore the error from close as there isn't anything we can do about it

Varun-Sethu (changes requested):
Minor changes otherwise looks good 🥳

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,89 @@ package repositories
...
+		return nil, err
+	}
+	return &fs, nil
+}
+
+// Add file to volume or update if exists
+func (c *DockerFileSystemRepository) AddToVolume(filename string) (err error) {
+	// Check if source file is valid
+	src, err := os.Open(filename)
+	defer src.Close()
```
Varun-Sethu, line 40: this should go after your error check

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,89 @@ package repositories
...
+	if err != nil {
+		return errors.New("Couldn't open source file")
+	}
+	// Create/update destination file and check it is valid
+	filepath := filepath.Join(volume_path, filename)
+	moved, err := os.OpenFile(filepath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
+	defer moved.Close()
```
Varun-Sethu, line 47: same here

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,89 @@ package repositories
...
+	if err != nil {
+		return errors.New("Couldn't remove the source file")
+	}
+	return nil
+}
+
+// Get file from volume. Returns a valid file pointer
+func (c *DockerFileSystemRepository) GetFromVolume(filename string) (fp *os.File, err error) {
+	// Concatenate volume path with file name
+	if file, err := os.Open(filepath.Join(volume_path, filename)); err != nil {
```
Varun-Sethu, line 67: fp, err = os.Open(filepath.Join(volume_path, filename))

Varun-Sethu:
```go
fp, err = os.Open(filepath.Join(volume_path, filename))
return
```

Backend/database/repositories/filesystem.go
```diff
@@ -3,14 +3,89 @@ package repositories
...
+	} else {
+		return file, nil
+	}
+}
+
+// Delete file from volume
+func (c *DockerFileSystemRepository) DeleteFromVolume(filename string) (err error) {
+	filepath := filepath.Join(volume_path, filename)
+	file, err := os.Open(filepath)
+	defer file.Close()
```
Varun-Sethu, line 78: same thing

Varun-Sethu (approved):

angary (approved):
LGTM!

alex-ljc (approved):

PR #148, opened 2022-06-06
[CMS-274] Adding api proxy for Development mode

fafnirZ:
### Why this change is necessary?
The old method of using `BACKEND_URI` does not work well in production environments and as a result we are using a custom api proxy server to proxy the requests to the `backend uri` in development mode

### Changes
- changed `/filesystems/*` to `/api/filesystems/*`
- installed dev dependency `http-proxy-middleware`
- added `setupProxy.ts` for proxy setup
- removed `${BACKEND_URI}` in CMS API Fetch requests
- added more makefile commands for selective building
- optimised frontend and backend Dockerfiles to be 90% faster to build
- downgraded nextjs compiled from `swc` to `babel`
- upgraded `node-alpine` version


Frontend/.dockerignore
```diff
@@ -4,3 +4,4 @@ build
 .dockerignore
 Dockerfile
 Dockerfile.prod
+DOckerfile.development
```
lhjt, line 7:
```suggestion
Dockerfile.development

```

Frontend/Dockerfile.development
```diff
@@ -0,0 +1,11 @@
+# pulling official base image
+FROM node:13.12.0-alpine
```
lhjt, line 2: Potentially consider updating this to the latest LTS?

lhjt (changes requested):

Frontend/Dockerfile.development
```diff
@@ -1,21 +1,11 @@
 # pulling official base image
-FROM node:13.12.0-alpine
+FROM node:18.2.0-alpine
```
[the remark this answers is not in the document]
lhjt: If it is working I'll mark this as resolved and approved.

lhjt (approved):

PR #153, opened 2022-06-12
Backend - MVP Pessimistic Concurrency Control Editor + Adding Published/Unpublished containers

Varun-Sethu:


Varun-Sethu:
## How to use these changes
This PR exposes a small TS client that you can use to interact with the server, the rest of this PR is just the serverside code. While the concurrent editor is in development this PR should server as an implementation of a simple **pessimistically** controlled editor. The idea is that each user acquires a lock to a document, if the lock cannot be acquired then they cannot edit the document, this is as opposed to our concurrent editor which implements OT to achieve **optimistic concurrency**  

The TS client can be used as follows:
```ts
let initDocumentCallback = (data: string) => globalState.data = data
let terminationCallback = (reason: TerminationReason) => globalState.AlertUser(arg)
let myDocumentID: number = 3

// create client
let client : Client = new Client(myDocumentID, initDocumentCallback, terminationCallBack)

// push changes to client (note these are queued if there are pending changes)
pushDocumentData("big chonky document state!")

```

fafnirZ (commented):
overall looks good, from my understanding the dockerfilesystems basically fopens and writes to a real new file in the docker volume, is this correct?

if so @lhjt is this potentially something we should have to look out for in terms of local file inclusion or RCE vulnerabilities?

Varun-Sethu:
> overall looks good, from my understanding the dockerfilesystems basically fopens and writes to a real new file in the docker volume, is this correct?
> 
> if so @lhjt is this potentially something we should have to look out for in terms of local file inclusion or RCE vulnerabilities?

There shouldnt be any local file inclusion or RCE vulnerabilities with this, all files within this volume are identified by a unqiue guid and only ever access via this guid, they can be thought of as a giant s3 bucket containing all the files really, the actually heirachy and organisation of these files is stored within the database. 

The backend never directly deals with a path from the user (at least one that maps to a real path on the physical OS its running on) so it should be fine.

fafnirZ (approved):
lgtm

PR #158, opened 2022-06-18
[CMS-236] Add image upload endpoint

angary:
A PR to handle the uploading of images, though have some questions:

- Should we make a new private function to creation and adding of entities to the fs, so that `CreateNewEntity` and `UploadImage` can call it, reducing additional code
- Should we also add logging to `ParseParamsToSchema`?

### DIFF
--- a/Backend/endpoints/filesystem_endpoints.go
+++ b/Backend/endpoints/filesystem_endpoints.go
@@ -216,6 +216,11 @@ func UploadImage(w http.ResponseWriter, r *http.Request, df DependencyFactory, l
 		return status, nil, nil
 	}
 
+	// Parse multipart file, with max size of 10 MB
+	var maxUploadSize int64 = 10 << 20
+	if err := r.ParseMultipartForm(maxUploadSize); err != nil {
+		return http.StatusBadRequest, nil, err
+	}
 	// Extract image and check for error
 	file, _, err := r.FormFile("Image")
 	if err != nil {
--- a/Backend/endpoints/filesystem_endpoints.go
+++ b/Backend/endpoints/filesystem_endpoints.go
@@ -219,14 +219,19 @@ func UploadImage(w http.ResponseWriter, r *http.Request, df DependencyFactory, l
 	// Extract image and check for error
 	file, _, err := r.FormFile("Image")
 	if err != nil {
-		fmt.Println("Error retrieving the file")
+		log.Write([]byte("Error retrieving the file"))
 		return http.StatusBadRequest, nil, err
 	}
 	defer file.Close()
 
 	// Create entity in repository
 	fs := reflect.TypeOf((*repositories.IFilesystemRepository)(nil))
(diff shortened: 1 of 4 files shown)


Backend/endpoints/filesystem_endpoints.go
```diff
@@ -203,3 +203,55 @@ func mapOver[T any, V any](input []T, mapFunction func(T) V) []V {
...
 	return output
 }
+
+type ValidImageUploadRequest struct {
+	Parent      int
+	LogicalName string `schema:"LogicalName,required"`
+	OwnerGroup  int    `schema:"OwnerGroup,required"`
+}
+
+func UploadImage(w http.ResponseWriter, r *http.Request, df DependencyFactory, log *logger.Log) (int, interface{}, error) {
```
Varun-Sethu, line 213: Do we want tests? :P

Backend/endpoints/util.go
```diff
@@ -41,6 +41,11 @@ func ParseParamsToSchema(r *http.Request, acceptingMethod string, target interfa
 		return http.StatusBadRequest
 	}
 
+	// Parse multipart file, with max size of 10 MB
+	if err = r.ParseMultipartForm(10 << 20); err != nil {
```
Varun-Sethu, line 45:
```go
var maxSize uint32 = 10 << 20 
if err = r.ParseMultipartForm(maxSize); err != nil {
    // ... do stuff
}
```

Backend/endpoints/util.go
```diff
@@ -41,6 +41,11 @@ func ParseParamsToSchema(r *http.Request, acceptingMethod string, target interfa
...
+		return http.StatusBadRequest
```
Varun-Sethu, line 46: The tests are failing here. HTTP requests without a multipart file attached to them will throw an error when you try and parse them. Perhaps we should have a separate method for parsing a schema with a multipart file

Backend/endpoints/filesystem_endpoints.go
```diff
@@ -203,3 +204,67 @@ func mapOver[T any, V any](input []T, mapFunction func(T) V) []V {
...
+		log.Write([]byte("Failed request!"))
+		return http.StatusNotAcceptable, nil, err
+	}
+
+	// Create and get a new entry in docker file system
+	dockerRepository.AddToVolume(strconv.Itoa(e.EntityID))
+	log.Write([]byte(fmt.Sprintf("Created new entity %v.", entityToCreate)))
+	dockerFile, err := dockerRepository.GetFromVolume(strconv.Itoa(e.EntityID))
+	if err != nil {
+		log.Write([]byte("Unable to get docker file"))
```
Varun-Sethu, line 255: Nitpick: You dont have a fullstop here but you have one everywhere else :P

Varun-Sethu (approved):

Backend/endpoints/filesystem_endpoints.go
angary: Guess I'll die

PR #160, opened 2022-06-19
Server State Helpers

brianzhang2101:
> Finished traversal and universal getter function. The getter now returns reflect.Value, due to some issues with how it was converted to interface{}

> Wrote unit tests for both functions although the getter largely relies on the validity of traversal

> Currently assuming the path is valid, have written some areas where error checks can be added in future. Also assuming requests keep the structure validated

> ParagraphAlign now correctly accepts specific String inputs

> Added unfinisted stub code for processing frontend requests based on the type of operation and parsing the data into the corresponding data types


Backend/editor/models/datamodel.go
```diff
@@ -1,38 +1,131 @@
...
+	Document_name string
+	Document_id   string
+	Content       []Component
 }
 
 // TODO:
 // - Add error checking for the paths as we traverse, e.g missing an index when traversing an array (assuming we didn't reach the end)
 // - Make sure the item we are adding keeps the validity of the object
 
 func process(request string) (err error) {
```
Varun-Sethu, line 27: I'm not really a fan of these named parameters tbh

Varun-Sethu: Should have pointed it out in the last PR :(

Backend/editor/models/datamodel.go
```diff
@@ -1,38 +1,131 @@
...
-	return nil, nil
+// TODO: Error check index logic
+
+// Gets the target object at the end of the path
+func (d Document) GetData(path string) (reflect.Value, error) {
+	paths, err := PathParser(path)
+	if err != nil {
+		return reflect.Value{}, err
+	}
+	parent := Traverse(d, paths)
```
Varun-Sethu, line 98: If we already have ```GetData``` does ```Traverse``` need to be exported? People should still already be able to use GetData for the same functionality right?

Backend/editor/models/paragraph.go
```diff
@@ -11,20 +14,25 @@ type Paragraph struct {
...
 
 // Get returns the reflect.Value corresponding to a specific field
-func (p Paragraph) Get(field string) (reflect.Value, error) {
+func (p *Paragraph) Get(field string) (reflect.Value, error) {
 	return reflect.ValueOf(p).FieldByName(field), nil
 }
 
 // Set sets a reflect.Value given a specific field
-func (p Paragraph) Set(field string, value reflect.Value) error {
+func (p *Paragraph) Set(field string, value reflect.Value) error {
```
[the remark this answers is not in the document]
Varun-Sethu: Also, should we calling .String more than once? I feel like .String is a rather expensive operation on a reflect.Value as all the processing happens at runtime. It would be better to compute value.String once and reuse it

Backend/editor/models/traversal.go
```diff
@@ -2,10 +2,52 @@ package models
...
+						if fieldType == reflect.Slice {
+							curr = field.Index(int(index))
+						} else {
+							curr = field.Elem().Index(int(index))
+						}
+					}
+				}
+				// Reflection returns a wrapper to a pointer to the interface of the struct
+				// Need to deference like so: interface -> ptr -> struct
+				if curr.Kind() == reflect.Interface {
```
Varun-Sethu, line 44: Can this not be in the switch too?

Varun-Sethu (changes requested):

Varun-Sethu:
Generally looks good! Just some needs some small changes here and there :D.

Varun-Sethu (approved):
Nice! :D

angary (approved):
LGTM!

Backend/editor/models/datamodel.go
```diff
@@ -1,38 +1,131 @@
...
 	payload Payload `json:"payload"`
 }
 
 type Document struct {
-	document_name string
-	document_id   string
-	content       []Component
+	Document_name string
+	Document_id   string
+	Content       []Component
```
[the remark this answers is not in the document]
brianzhang2101: This is already fixed buddy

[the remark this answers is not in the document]
angary: Idk why this comment appeared again

angary:
Un-LGTM'ed should use `strings.Atoi(subpaths[i])` instead of `strconv.ParseInt(subpaths[i], 10, 32)` so that you don't have to typecast from `int64` to an `int`

brianzhang2101:
bruh you couldve just fixed this locally...

PR #163, opened 2022-07-03
[CMS-287] added small reusable container

leannachan:
Preview of the small reusable container
image


Next/src/pages/container/blog.css
```diff
@@ -0,0 +1,10 @@
+/* image 1 */
+
+position: absolute;
+width: 340px;
+height: 237px;
+left: 149px;
+top: 945px;
+
+background: url(image.png);
+border-radius: 50px;
```
fafnirZ, line 10: we dont use .css files in our project, please remove and try use other methods

Next/src/mock/index.ts
```diff
@@ -45,3 +46,28 @@ export const data: Element[] = [
...
+      {text: "Optimise Your Study Life:Study Apps", bold: true}
+    ]
+  },
+  {
+    type: "paragraph",
+    children: [
+      {text: "In the second part of the series, Alex Xu walks us through study apps you can use to boost your productivity!"}
+    ]
+  }
+]
```
fafnirZ, line 73: is this the most efficient way to present the data? 

Firstly, having an array represent sections of 1 container doesnt make that much sense since at a top level, 1 object should represent 1 container, and all sections should be nested inside the container object as sections

Secondly, is it necessary to have the children tag representing the content of an object
For example:
```
{
  type: "title",
  text: "articles"
}
```
is sufficient, and similarly with everything else, 
there is no need to create more un-necessary nesting

Thirdly, I dont think for your container, it is not necessary to introduce section block mapping, i.e. utilising the `map` function to map out individual sections.

fafnirZ (changes requested):
Please Address the changes, I think there should be a re-architecture of your logic, since there is a lot of over engineering, the task at hand is quite simple try to keep it simple

fafnirZ:
Closing PR due to huge backlog of unmerged changes, please manually create new branch from *updated* main branch before proceeding with migrating changes to the new updated branch

PR #168, opened 2022-07-16
Fixed State Sending from Backend

brianzhang2101:
- Fixed file reset to blank after re-entering editor. Data is now persistent (can be verified within Docker desktop, under unpublished data volume)
- Fixed buffer not reading file content properly
- Fixed JSON message from backend incorrectly using quotation marks when expecting an object. Empty files will now send an empty array
- Fixed conflicting dependency version issue with "react-refresh". Previous developers who had frontend working will not have experienced this, but fresh installs will run into this issue


Varun-Sethu (approved):
LGTM! 😎

zijizhu (approved):
Looks good!

PR #169, opened 2022-07-17
[CMS-275] Editor Publish Button

zijizhu:
- Backend fixes for editor connection from #168 
- Implemented a publish button that syncs the updated content with backend


Frontend/package.json
```diff
@@ -90,5 +90,8 @@
     "eslint-plugin-storybook": "0.5.12",
     "typescript": "4.6.2",
     "webpack": "5.70.0"
+  },
+  "overrides": {
+    "react-refresh": "0.11.0"
```
fafnirZ, line 95: what is this for?

fafnirZ (approved):
lgtm, where in the nextjs repo does the webpage consume these published updates? is that feature connected yet? We need it to be connected to be working E2E

Frontend/package.json
zijizhu: > what is this for?

This is from #168:

> Fixed conflicting dependency version issue with "react-refresh". Previous developers who had frontend working will not have experienced this, but fresh installs will run into this issue

PR #171, opened 2022-07-19
[CMS-296] Add popup functionality for Content Block Popup UI

lauraw0:

- Needed popup functionality for a content block that manages uploads

## **Changes**

- Implemented popup with a transition and transparent background

## **Screenshots**
Before clicking button:
image
After clicking button:
image


fafnirZ (approved):
please rename files, lets keep it to using camelCase, no snake cases in our codebase

PR #173, opened 2022-07-25
Fix edge case in traversal and add more test cases for invalid inputs

angary:
- Change to use `strconv.Atoi` instead of `strconv.ParseInt`
- Add bonus test cases for invalid case
- Return error if the ascii to integer fails (if there was an error, index would be 0) resulting in invalid field)

Also wondering if there is a way to do deep equality checks - this would make the traversal testing less rigid, i.e. we check that the struct returned from traversal has deep equality with struct in `setupDocument`

### DIFF
--- a/Backend/editor/models/tests/models_test.go
+++ b/Backend/editor/models/tests/models_test.go
@@ -73,8 +73,7 @@ func TestInvalidFieldName(t *testing.T) {
 	if err != nil {
 		log.Fatalf(err.Error())
 	}
-	result, err := models.Traverse(testObj, subpaths)
-	_ = result // Stop go from complaining about unused variable
+	_, err = models.Traverse(testObj, subpaths)
 	assert.NotNil(t, err)
 }
 
@@ -101,8 +100,7 @@ func TestNonIntegerArrayIndex(t *testing.T) {
 	if err != nil {
 		log.Fatalf(err.Error())
 	}
-	result, err := models.Traverse(testObj, subpaths)
-	_ = result // Stop go from complaining about unused variable
+	result, err = models.Traverse(testObj, subpaths)
 	assert.NotNil(t, err)
 }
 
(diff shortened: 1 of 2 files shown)


Varun-Sethu:
To answer your question, IIRC there is a `reflect.DeepEqual`

Backend/editor/models/tests/models_test.go
```diff
@@ -79,7 +66,19 @@ func TestSliceField(t *testing.T) {
...
 
-func TestArrayField(t *testing.T) {
+func TestInvalidFieldName(t *testing.T) {
+	testObj := setupDocument()
+	path := "Content/0/InvalidFieldName/0"
+	subpaths, err := models.ParsePath(path)
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+	result, err := models.Traverse(testObj, subpaths)
```
Varun-Sethu, line 76:
```go
_, err := models.Traverse(testObj, subpaths)
```

Backend/editor/models/tests/models_test.go
```diff
@@ -95,7 +94,31 @@ func TestArrayField(t *testing.T) {
...
 
-func TestStructField(t *testing.T) {
+func TestNonIntegerArrayIndex(t *testing.T) {
+	testObj := setupDocument()
+	path := "Content/asdf/Data"
+	subpaths, err := models.ParsePath(path)
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+	result, err := models.Traverse(testObj, subpaths)
```
Varun-Sethu, line 104: Same here

Backend/editor/models/tests/models_test.go
```diff
@@ -95,7 +94,31 @@ func TestArrayField(t *testing.T) {
...
+}
+
+func TestOutOfBoundsArrayIndex(t *testing.T) {
+	testObj := setupDocument()
+	path := "Content/4/Data"
+	subpaths, err := models.ParsePath(path)
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+	result, err := models.Traverse(testObj, subpaths)
```
Varun-Sethu, line 116: and here

Varun-Sethu (approved):

brianzhang2101 (approved):
lgtm

PR #175, opened 2022-07-26
Editor DataModel Refactor

Varun-Sethu:
This PR just refactors the data-model folder a little, it generalizes a lot of the existing functions so they can work on arbitrary data-types and also cleans up some code.


Backend/editor/data/jsonConfig.go
```diff
@@ -0,0 +1,55 @@
...
+	"strconv"
+
+	"cms.csesoc.unsw.edu.au/editor/data/datamodels/cmsmodel"
+	"cms.csesoc.unsw.edu.au/internal/cmsjson"
+)
+
+// models contains all the data models for the editor
+// this is the configuration required for the cmsjson module
+// cmsjson is a custom marshaller/unmarshaller that supports interface types
+// cmsjson works with arbtirary schemas so this model can be changed on a whim
```
angary, line 15:
```suggestion
// cmsjson works with arbitrary schemas so this model can be changed on a whim
```

Backend/editor/data/applicator.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"cms.csesoc.unsw.edu.au/editor/data/datamodels"
+)
+
+// TODO:
+// - Add error checking for the paths as we traverse, e.g missing an index when traversing an array (assuming we didn't reach the end)
+// - Make sure the item we are adding keeps the validity of the object
+
+// ApplyRequest takes a datamodel (as defined in the datamodels folder) and a request, it then proceeds to apply the request
+// to the model, note that this assumes that the oepration in the request has been appropriately transformed
```
angary, line 17:
```suggestion
// to the model, note that this assumes that the operation in the request has been appropriately transformed
```

Backend/internal/cmsjson/marshall.go
```diff
@@ -107,13 +107,14 @@ func (c Configuration) marshallPrimitive(field reflect.Value) []byte {
 func (c Configuration) marshallCore(source reflect.Value) []byte {
```
angary, line 107: Maybe use switch statements for the function above this (`marshallPrimitive`)?

Backend/editor/data/jsonConfig.go
```diff
@@ -0,0 +1,55 @@
...
+		reflect.TypeOf((*Payload)(nil)).Elem(): {
+			"textEdit":  reflect.TypeOf(TextEdit{}),
+			"keyEdit":   reflect.TypeOf(KeyEdit{}),
+			"arrayEdit": reflect.TypeOf(ArrayEdit{}),
+		},
+	},
+}
+
+// small helper function to parse a JSON value of a specific type
+func parseDataGivenType(dataStr string, dataType string) (interface{}, error) {
```
angary, line 36: Not sure when this function is used? Would it be be better to just use a function in the `unmarshall.go` file by making that function public, or is this for something different / we need this function because inputs are different?

Backend/internal/cmsjson/marshall.go
angary: I can't seem to leave a comment on that piece of code since the code wasn't changed in this PR.

angary (changes requested):
Almost LGTM! Very nice touch with the `FieldByName`

Backend/editor/data/jsonConfig.go
Varun-Sethu: I'm not sure either, Brian had it there initially so I'm assuming he plans on using it soon.

Backend/editor/data/applicator.go
```diff
@@ -0,0 +1,44 @@
...
+
+// TODO:
+// - Add error checking for the paths as we traverse, e.g missing an index when traversing an array (assuming we didn't reach the end)
+// - Make sure the item we are adding keeps the validity of the object
+
+// ApplyRequest takes a datamodel (as defined in the datamodels folder) and a request, it then proceeds to apply the request
+// to the model, note that this assumes that the operation in the request has been appropriately transformed
+func ApplyRequest(model datamodels.DataModel, request Request) error {
+	_, err := GetOperationTargetSite(model, request.path)
+	if err != nil {
```
angary, line 20: Could be 1 line, up to you :))

Backend/editor/data/traversal.go
```diff
@@ -0,0 +1,84 @@
...
+		if !prev.IsValid() || !curr.IsValid() {
+			return reflect.Value{}, reflect.Value{}, fmt.Errorf("invalid path, couldn't find subpath %s", subpaths[0])
+		}
+	}
+
+	return prev, curr, nil
+}
+
+// tryConsumeArrayElement attempts to consume the head of the path, the assumption is that the current value is an
+// array value, if this is not the case then the function returns the prev, current and path uncahnged, if it was capable
```
angary, line 29:
```suggestion
// array value, if this is not the case then the function returns the prev, current and path unchanged, if it was capable
```

Backend/editor/data/traversal.go
```diff
@@ -0,0 +1,84 @@
...
+func Traverse(document datamodels.DataModel, subpaths []string) (reflect.Value, reflect.Value, error) {
+	prev := reflect.Value{}
+	curr := reflect.ValueOf(document)
+
+	for len(subpaths) > 0 {
+		prev, curr, subpaths = consumeField(prev, curr, subpaths)
+		prev, curr, subpaths = tryConsumeArrayElement(prev, curr, subpaths)
+
+		if !prev.IsValid() || !curr.IsValid() {
+			return reflect.Value{}, reflect.Value{}, fmt.Errorf("invalid path, couldn't find subpath %s", subpaths[0])
```
angary, line 21: A couple error messages were removed in this refactor so would be useful to have more information here for debugging / logging, esp if there's multiple subpaths with the same name. 

Maybe make a copy of subpaths like `remainingSubpath` and which is consumed by the above "consume" functions, and then you can do something like 

```go
for len(remainingSubpath) > 0 {
    // consume
    if someError {
        currIndex := len(subpaths) - len(remainingSubpaths) // maybe this is 1 off idk
        err := fmt.Error("invalid path, couldn't find subpath %s", strings.Join(subpaths[:currIndex], "/");
        return ..., err
}

File: Backend/editor/data/traversal.go [hunk 8]
```diff
@@ -0,0 +1,84 @@
...
+func consumeField(prev, curr reflect.Value, path []string) (reflect.Value, reflect.Value, []string) {
+	field := curr.FieldByName(path[0])
+	if curr.IsValid() {
+		return curr, dereference(field), path[1:]
+	}
+
+	return prev, reflect.Value{}, path
+}
+
+// getValueAtIndex fetches the value at the provided index in the array pointed at by reflect.Value
```
angary, line 64:
```suggestion
// getValueAtIndex fetches the value at the provided index in the array or slice pointed at by reflect.Value
```

Backend/editor/data/traversal.go
```diff
@@ -0,0 +1,84 @@
...
+// note it is assumed that index is a string
+func getValueAtIndex(array reflect.Value, i string) reflect.Value {
+	index, err := strconv.Atoi(i)
+	if err != nil || index >= array.Len() || index < 0 {
+		return reflect.Value{}
+	}
+
+	return array.Index(index)
```
angary, line 72: You may have to check if it's an array or a slice, since one requires `array.Elem().Index(index)` iirc. Maybe also change variable name from `array` to `list` or `l` or `xs` if you wanna be epic haskell man.

angary (changes requested):

Backend/editor/data/traversal.go
Varun-Sethu: No Haskell >:( , also the .Elem is if/when u allocate a array/slice with new, we're not really allocating anything so .Index works for both

Backend/editor/data/applicator.go
Varun-Sethu: This is temporary placeholder function, when actually implemented the ignored variable will be "unignored" so best to keep it as is

angary (approved):
LGTM!

Backend/editor/data/jsonConfig.go
brianzhang2101: This function was made on the original specification of taking in requests from frontend. Each request would pass in the payload with content and content type. Not sure if things have changed since I made this months ago?

brianzhang2101: It's not currently in use, but was meant to be linked to the entrypoint once it was done. @Varun-Sethu thoughts?

Backend/editor/data/applicator.go
```diff
@@ -0,0 +1,44 @@
...
+	default:
+		return errors.New("invalid edit type")
+	}
+
+	return nil
+}
+
+// getOperationTargetSite Gets the target object at the end of the path, this is the operation that we need to apply
+// our operation to
+func GetOperationTargetSite(model datamodels.DataModel, subpaths []string) (reflect.Value, error) {
```
brianzhang2101, line 37: Any reason why the original getter was removed and replaced with a wrapper to Traversal?

brianzhang2101 (approved):
super clever refactoring, LGTM

PR #176, opened 2022-07-30
CMS251 - Adding OT Transform functions

alex-ljc:


Backend/editor/tests/transform_tests.go
```diff
@@ -0,0 +1,136 @@
+package editor
```
Varun-Sethu, line 1: even though this says: "package editor" the package it's not really in the same package as your functions, this is the package: `cms.csesoc.unsw.edu.au/editor/tests/` while your functions are in `cms.csesoc.unsw.edu.au/editor` hence why you are getting import errors, you're trying to access private functions (functions that start with lower case letters) from an external package

Backend/editor/tests/transform_tests.go
```diff
@@ -0,0 +1,136 @@
...
+	var pos3 = []int{1, 0}
+	var pos4 = []int{1, 0, 3, 2}
+	assert.Equal(t, transformPoint(pos3, pos4), 1)
+}
+
+func TestEffectIndependence(t *testing.T) {
+	// Effect independence condition 1 (refer to https://arxiv.org/pdf/1512.05949.pdf)
+	var pos1 = []int{2, 2, 3, 4}
+	var pos2 = []int{1, 2, 4, 4}
+	assert.Equal(t, effectIndependent(pos1, pos2, transformPoint(pos1, pos2)), true)
```
Varun-Sethu, line 23: either export these functions in transform.go by capitalising them (not ideal) or move your transform_test.go file to the same directory as transform.do

PR #200, opened 2022-08-28
Change filesystem entity ids to be uuids

angary:
- Also remove `rootroot`
- `root` now has entity id of `uuid_nil()` and it's parent is `uuid_nil()` (points to itself) because the postgres driver has issue supporting `NULL` values


postgres/create_tables.sql
```diff
@@ -145,13 +145,13 @@ DECLARE
   wasPopping    filesystem.EntityID%type;
   oldEntity     filesystem.EntityID%type;
 BEGIN
-  SELECT filesystem.EntityID INTO rootID FROM filesystem WHERE Parent = 0;
+  SELECT filesystem.EntityID INTO rootID FROM filesystem WHERE EntityID = uuid_nil();
   
-  newEntity := (SELECT new_entity(2, 'downloads'::VARCHAR, 1, false));
```
angary, line 150: Double check the logic here, I wasn't too sure what the magic number 2 meant, and what the purpose of this was

backend/database/repositories/main.go
```diff
@@ -15,7 +15,7 @@ const (
 )
 
 // The ID for root, set this as the ID in a specified request
-const FILESYSTEM_ROOT_ID = 0
+const FILESYSTEM_ROOT_ID string = "00000000-0000-0000-0000-000000000000"
```
Varun-Sethu, line 18: I'll add a general comment later but I think this is a good example, IMO it would be better if we surfaced these primary keys are UUID types as opposed to strings, checkout this library for a nice uuid type + relevant operations: https://github.com/google/uuid

but by using this library we can turn this line into:
```go
const FILESYSTEM_ROOT_ID uuid.UUID = uuid.Nil 
```

backend/database/repositories/filesystem.go
```diff
@@ -14,7 +14,7 @@ type filesystemRepository struct {
 // We really should use an ORM jesus this is ugly
 func (rep filesystemRepository) query(query string, input ...interface{}) (FilesystemEntry, error) {
 	entity := FilesystemEntry{}
-	children := []int{}
+	children := []string{}
```
Varun-Sethu, line 17: Following on from my comment in `repositories/main.go` if we go with an actual UUID type then this would be turned into:

```go
children := []uuid.UUID

for rows.Next() {
   var x string
   rows.Scan(&x)

   if parsedId, err := uuid.Parse(x); err != nil {
     children = append(children, parsedId)
   } else {
     return FilesystemEntry{}, err
   }
}
```

Varun-Sethu (changes requested):
In general I feel like it would be better if everything was the concrete `uuid.UUID` type as opposed to a string, doing this allows us to manipulate EntityPKs just like any other UUID, see my comment in `database/repositories/main.go` for more details

backend/database/repositories/filesystem.go
angary: Currently done something like 

```go
    children := []uuid.UUID{}
	// finally scan in the rows
	for rows.Next() {
		var x uuid.UUID
		err := rows.Scan(&x)
		if err != nil {
			return FilesystemEntry{}, err
		}

		children = append(children, x)
	}
```
i.e. changed `x` to be of type `uuid.UUID` and did not include `uuid.Parse`

angary:
Also the filenames are string forms of uuids, so I've left functions like `GetFromVolume(filename)` take in a type string rather than uuids

Varun-Sethu (approved):
LGTM! 😎

PR #202, opened 2022-08-30
implemented (did not get chance to test) part of the authentication

alex-ljc:


backend/database/repositories/filesystem.go
```diff
@@ -21,13 +22,15 @@ func (rep filesystemRepository) query(query string, input ...interface{}) (Files
 		&entity.EntityID, &entity.LogicalName, &entity.IsDocument, &entity.IsPublished,
 		&entity.CreatedAt, &entity.OwnerUserId, &entity.ParentFileID)
 	if err != nil {
+		fmt.Print(err.Error())
```
Varun-Sethu, line 25: Sorry I left this in while testing, please undo

backend/endpoints/dependency_factory.go
```diff
@@ -80,3 +82,8 @@ func getDependency[T any](factory DependencyFactory) T {
 
 	return factory.GetDependency(factory.GetDepFromType(typeMapping)).(T)
 }
+
+func setFrontEndID(id int, df interface{}) {
+	stype := reflect.ValueOf(df).Elem()
+	stype.FieldByName("FrontEndID").SetInt(int64(id))
+}
```
Varun-Sethu, line 89: 1. What is this for?
2. Please don't use reflection where its not needed
3. This only works on the very specific implementation of DF which is our `dependencyProvider`, it will panic on everything else

backend/endpoints/main.go
```diff
@@ -65,7 +67,13 @@ func (fn handler[T, V]) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// construct a dependency factory for this request, which implies instantiating a logger
 	logger := buildLogger(r.Method, r.URL.Path)
-	dependencyFactory := DependencyProvider{Log: logger}
+	var frontendid int
+	ctx := contexts.GetDatabaseContext()
+	err := ctx.Query("SELECT frontendid from frontend where frontendurl = $1;", []interface{}{r.URL.Path}, &frontendid)
+	if err != nil {
+		log.Println("frontend doesn't exist", err.Error())
+	}
```
Varun-Sethu, line 75: I don't like the idea of querying the database directly from the endpoint package, this should be in the repositories package with its own repository type, IE:

```go
type (
    IFrontendRepository interface {
        GetFrontendFromURL(url string) int
    }
)
```

We can then acquire and call the repository as follows:
```go
frontendRepo := repositories.GetRepository(FRONTEND).(IFrontendRepository)
frontendID := frontendRepo.GetFrontendFromURL(r.URL.Path)
```

Varun-Sethu: Also the logger being used is incorrect, endpoints have their own custom logger type so we can easilly group together relevant log messages without them being tangled together in a mess, please use the logger defined on line 69

backend/database/repositories/person.go
```diff
@@ -11,11 +11,12 @@ import (
 // Implements IPersonRepository
 type personRepository struct {
 	embeddedContext
+	FrontEndID int
```
Varun-Sethu, line 14: nitpicks:
 - Swap order of fields
 - FrontEndID should not be exported (it should start with a lowercase letter)

backend/database/repositories/main.go
```diff
@@ -28,10 +30,6 @@ func GetRepository(repo int) interface{} {
 		return filesystemRepository{
 			embeddedContext{context},
 		}
-	case PERSON:
-		return personRepository{
-			embeddedContext{context},
-		}
```
Varun-Sethu, line 34: You removed this from the GetRepository function however the dependency_factor still attempts to instantiate it (line 31 of `dependency_factory.go`), please remove that reference otherwise we could very easilly have a nil pointer dereference

Varun-Sethu (changes requested):

Varun-Sethu (approved):

PR #232, opened 2022-09-16
feat(cms): add string transformation functions

angary:
- Implement string transformation functions
- Change `type Operation` to `interface Operation` with `transformAgainst` method
- Fix up `transform` function to include atomic operation
- Split `operation.ts` file to `operation.ts` (purely `Operation` and `AtomicOperation` classes + methods and `transform.ts` (purely path transformation function for `Operation`
- Add tests in `transform.test.ts` and `operation.test.ts`, though they're there to ensure that the string operations work atm, more can be added later


backend/editor/OT/data/string_operation.go
```diff
@@ -1,20 +1,112 @@
...
+		return o1
+	} else if o1.RangeStart >= o2.RangeEnd {
+		length := o2.RangeEnd - o2.RangeStart
+		o1.RangeStart, o1.RangeEnd = o1.RangeStart-length, o1.RangeEnd-length
+	} else {
+		if o2.RangeStart <= o1.RangeStart {
+			// o2 starts before or at o1
+			if o1.RangeEnd <= o2.RangeEnd {
+				// o2 is bigger or same as o1 so o1 becomes noop
+				o1.RangeEnd = o1.RangeStart
```
Varun-Sethu, line 95: In hindsight I think the way we've set this whole thing up is rather hacky, I think `IsNoOp` should be a property of the OperationModel itself as opposed to the Operation, probably something to clean up after the mvp

backend/editor/OT/data/tests/string_operation_test.go
```diff
@@ -0,0 +1,180 @@
...
+	o2 := data.StringOperation{RangeStart: 2, RangeEnd: 3, NewValue: "2"}
+	o1_t1, o2_t1 := o1.TransformAgainst(o2, data.Insert)
+	o1_t2, o2_t2 := o2.TransformAgainst(o1, data.Insert)
+
+	assert := assert.New(t)
+	assert.Equal("a1b2cde", apply(o1_t1, apply(o2_t1, s)))
+	assert.Equal(apply(o1_t1, apply(o2_t1, s)), apply(o2_t2, apply(o1_t2, s)))
+}
+
+// TODO: In the case that they are the same location, we need vector clock / ID to determine which operation is done first
```
Varun-Sethu, line 24: See my comments on discord about this

Varun-Sethu (approved):
Nice! 😎

backend/editor/OT/OTClient/operation.ts
```diff
@@ -1,165 +1,252 @@
...
+  (a[tp] < b[tp] && a.length > b.length);
+
+/**
+ * Normalise turns an empty operation into a noop
+ */
+const normalise = (a: Operation): Operation => (a.path.length === 0 ? noOp : a);
+
+const copy = <T>(a: T): T => JSON.parse(JSON.stringify(a));
+
+export const stringTransform = (
```
Varun-Sethu, line 172: So the general idea behind these type specific transformation functions is that we invoke these transformation functions whenever our operations don't morph the document tree but could still result in conflicts (at the level of individual properties), I think it might be worthwhile doing a simmilar thing to what we do in Go where we define an EditType interface which contains this Transform funtion and then invoke that whenever we run into the above situation

Varun-Sethu: I.e we have something like:
```ts
interface AtomicOperation {
   $type: string;
   transformAgainst: (op: AtomicOperation) => {AtomicOperation, AtomicOperation}
   // We leave out apply as our client is only concerned with transformation not application
}

class StringOperation implements AtomicOperation {
   $type: string = "stringOperation";
   transformAgainst = (op: AtomicOperation) => {
    // your string transformation code
   }
}
```
So then within our actual general transformation functions once we detect a scenario where the conflict is at the level of invidual values and not the document tree we can invoke:
```ts
Operation.operation.transformAgainst(...)
```
On a sidenote, these properties should probably have better names :P `Operation.operation` is rather odd

Varun-Sethu (changes requested):
Overall looks good, just some smol changes needed

backend/editor/OT/OTClient/operation.ts
angary: nah it's good to me

Varun-Sethu (approved):

PR #259, opened 2022-09-25
feat(cms): Support sending raw image and json in `GetPublishedDocument` endpoint

angary:


backend/endpoints/models/volume_models.go
```diff
@@ -32,6 +32,6 @@ type (
 type (
 	// DocumentRetrievalResponse is just the returned response for any handler that fetches the contents of a docker volume
 	DocumentRetrievalResponse struct {
```
Varun-Sethu, line 34: Do we still need this type? The output type of `GetPublishedDocument` is now just an array of bytes as opposed to a nicely formatted json response

Varun-Sethu (approved):

backend/endpoints/models/volume_models.go
angary: Good catch removed now

PR #263, opened 2022-09-26
Database Migration Tool

Varun-Sethu:
This is an absolutely quick and hacky database migrations, it deletes all tables and rebuilds them, in the future we will use a proper migration tool but this should suffice to get staging up and running


postgres/migrate.py
```diff
@@ -0,0 +1,98 @@
...
+    connection.autocommit = False
+    return connection
+
+
+# UpgradeJob represents a single attempt to upgrade the DB, it does everything within a transaction
+class UpgradeJob:
+    def __init__(self):
+        self.connection = get_db()
+        self.cursor = self.connection.cursor()
+        pass
```
angary, line 28: ???????????????

angary (approved):
unLGTM, try not to do `except: pass`

postgres/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:latest
+
+COPY . .
+RUN pip install -r requirements.txt
+
+ENTRYPOINT [ "python", "migrate.py" ]
```
lhjt, line 6: I'll have to optimise this file a bit before this PR can be merged.

lhjt (commented):
I'll have to push a few changes so that this image is created by the CI pipelines. Just a few comments to look at; I'll merge this in once I've added the relevant CI changes.

docker-compose.yml
```diff
@@ -52,8 +55,21 @@ services:
...
+      - db
+    environment:
+      - POSTGRES_HOST=db
+      - POSTGRES_DB=${PG_DB}
+      - POSTGRES_USER=${PG_USER}
+      - POSTGRES_PASSWORD=${PG_PASSWORD}
+
   staging_db:
     container_name: pg_container_testing
     image: postgres
```
lhjt, line 75: What's the purpose of this db? (compared to the other `db` service in this file)

Varun-Sethu: It was/is an ease of use DB we have to run any database bound tests against, although we already wrap all our tests in transactions so its not entirely required, ill remove it within another PR if need be

lhjt (approved):
Looks good; I'll let you know once I edit the deployment manifests so that this is live.

PR #294, opened 2022-10-18
[CMS-306] Thomas L - Onboarding

ShunyaoLiang:
Hello!

This adds tests for `TestLogin` and `TestLogout` in [`auth_endpoints.go`](https://github.com/csesoc/website/blob/CMS-306/backend/endpoints/auth_endpoints.go). Sorry for taking so long.

They test only the current behaviour on the typical success case, that is, logging in as a user that exists and logging out a user that has logged in. There is definitely space to add tests for error cases.

Part of the success behaviour is that `TestLogin` returns the status code [`301 Moved Permanently`](https://http.cat/301), which seems dubious to me. That seems more appropriate for deprecated routes, especially since RFC 2616 says that it indicates the client should update references to point to the redirect and that the response is cacheable.

Lastly, I exported `(u *User) HashPassword` (previously `(u *User) hashPassword`) from `endpoints.models`. I thought how the tests generated entries in the person repository should be coupled to how the implementation does it.

### DIFF
--- a/backend/endpoints/models/user_models.go
+++ b/backend/endpoints/models/user_models.go
@@ -32,7 +32,7 @@ func (u *User) IsValidEmail() bool {
 
 // UserExists just checks if a user has a valid password
 func (u *User) UserExists(personRepo repositories.PersonRepository) bool {
-	hashedPassword := u.hashPassword()
+	hashedPassword := u.HashPassword()
 
 	return personRepo.PersonExists(repositories.Person{
 		Email:    u.Email,
@@ -40,8 +40,8 @@ func (u *User) UserExists(personRepo repositories.PersonRepository) bool {
 	})
 }
 
-// hashPassword hashes a user's password
-func (u *User) hashPassword() string {
+// HashPassword hashes a user's password
+func (u *User) HashPassword() string {
 	hashedBytes := sha256.Sum256([]byte(u.Password))
 	return string(hashedBytes[:])
 }
--- a/backend/endpoints/tests/auth_test.go
+++ b/backend/endpoints/tests/auth_test.go
@@ -16,22 +16,19 @@ import (
 	. "cms.csesoc.unsw.edu.au/endpoints/mocks"
 )
 
+const TEST_EMAIL = "thomas.liang1@student.unsw.edu.au"
+const TEST_PASSWORD = "backendnumberone!"
+
 // Test [endpoints.LoginHandler] succeeds on correct input.
 func TestLogin(t *testing.T) {
-	form := models.User {Email: "thomas.liang1@student.unsw.edu.au", Password: "backendnumberone"}
+	form := newTestUser()
 	responseRecorder := httptest.NewRecorder()
 	request := httptest.NewRequest("POST", "/login", nil)
 
 	// Set up mock-test boilerplate.
 	controller := gomock.NewController(t)
 	defer controller.Finish()
-	// LoginHandler queries the person repository of the dependency factory it is given for the user details.
-	mockPersonRepository := NewMockIPersonRepository(controller)
-	// Fake a repository entry.
-	person := repositories.Person {Email: form.Email, Password: form.HashPassword()}
-	mockPersonRepository.EXPECT().PersonExists(person).Return(true)
-	mockDependencyFactory := NewMockDependencyFactory(controller)
-	mockDependencyFactory.EXPECT().GetPersonsRepo().Return(mockPersonRepository)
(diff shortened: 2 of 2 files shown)


backend/endpoints/models/user_models.go
```diff
@@ -32,16 +32,16 @@ func (u *User) IsValidEmail() bool {
...
 	return personRepo.PersonExists(repositories.Person{
 		Email:    u.Email,
 		Password: hashedPassword,
 	})
 }
 
-// hashPassword hashes a user's password
-func (u *User) hashPassword() string {
+// HashPassword hashes a user's password
+func (u *User) HashPassword() string {
```
Varun-Sethu, line 44: I kind of don't like the idea of exporting this just so it can be used by tests, my main gripe with the idea is that it suddenly allows for someone to completely change the hash algorithm and tests would still pass, ie redefining `HashPassword` to
```go
func (u *User) HashPassword() string {
    return u.Password
}
```
would still pass all the tests but would still be a problematic change :(

backend/endpoints/tests/auth_test.go
```diff
@@ -1,11 +1,93 @@
...
+	. "cms.csesoc.unsw.edu.au/endpoints/mocks"
+)
+
+const TEST_EMAIL = "thomas.liang1@student.unsw.edu.au"
+const TEST_PASSWORD = "backendnumberone!"
+
+// Test [endpoints.LoginHandler] succeeds on correct input.
+func TestLogin(t *testing.T) {
+	form := newTestUser()
+	responseRecorder := httptest.NewRecorder()
```
Varun-Sethu, line 25: nice, TIL this was a thing :P

backend/endpoints/tests/auth_test.go
```diff
@@ -1,11 +1,93 @@
...
+
+	// Set up mock-test boilerplate.
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+	mockDependencyFactory := setUpMockRepositories(controller)
+
+	response := endpoints.LoginHandler(form, responseRecorder, request, mockDependencyFactory)
+
+	// The fun happens here!
+	const statusCode = http.StatusMovedPermanently
```
Varun-Sethu, line 36: Yeah I agree with your comments about ```http.StatusMovedPermanently``` I think it may have been added because that code was simply copy and pasted from SO? IMO our backend really shouldn't be redirecting to the dashboard page, I feel like that should be something that the frontend does instead, perhaps its best to remove that entire redirection line entirely

backend/endpoints/models/user_models.go
ShunyaoLiang: Hmm that makes a lot of sense.

With the current implementation, it is necessary that the tests know how passwords are hashed to mock the person repository. The only way of avoiding exporting `hashPassword` I can think of is adding a hash function to `IPersonRepository`. Thoughts?

To that last remark, I think it's also worth considering if it is the responsibility of these tests to check if the hashing algorithm is problematic; that seems like a code review issue 😛

backend/endpoints/tests/auth_test.go
ShunyaoLiang: Thanks! Lowkey sick of the signature Go documentation teal already 😩

backend/endpoints/tests/auth_test.go
ShunyaoLiang: I see. I'm not really sure what I should do about this.

Varun-Sethu: Hmm, would you perhaps mind if we created a separate ticket to branch off this? (Sorry for the delay btw)

backend/endpoints/models/user_models.go
Varun-Sethu: IMO unit tests should assert behavior and I feel like the hashing algorithm falls under that, although you do raise a good point about moving it to the `IPersonRepository`, actually moving the hashing logic feels like a better place than within the endpoint logic. I guess for now we can leave it as it is and spawn an offshoot ticket for updating the `IPersonRepository`?

Varun-Sethu (approved):

backend/endpoints/models/user_models.go
ShunyaoLiang: Sounds good to me.

backend/endpoints/tests/auth_test.go
ShunyaoLiang: Sounds good to me.

PR #298, opened 2022-10-31
OT Frontend Changes - Stage 1

Varun-Sethu:
PR adds the first stage of the required changes to the front-end in order to support operational transform. This PR captures SlateJS operations and propagates them to a central place for delivery to the server. The PR after this will focus on translating SlateJS operations into the CMS operation format + sending them to the server and the final PR will focus on applying operations sent from the server to the editor.

### DIFF
--- a/postgres/Dockerfile
+++ b/postgres/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-alpine
+FROM python:alpine
 
 COPY requirements.txt .
 RUN apk update && apk add postgresql-dev gcc python3-dev musl-dev
--- a/postgres/Dockerfile
+++ b/postgres/Dockerfile
@@ -1,6 +1,7 @@
-FROM python:slim
+FROM python:3.6-alpine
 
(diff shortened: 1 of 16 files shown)


Varun-Sethu:
Tests are failing because there appears to be something wrong with PIP implying the DB startup script couldn't be run and consequentially the backend testing env couldn't be created

zijizhu (approved):
Looks good to me!

lauraw0 (approved):
lgtm! :)

postgres/Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM python:slim
+FROM python:3.6-alpine
```
lhjt, line 1: Any reason why it must be this old?

lhjt (changes requested):

postgres/Dockerfile
Varun-Sethu: should be fixed now

lhjt (approved):

PR #301, opened 2022-11-09
chore(cms): Setting up OT editor for integration testing

Varun-Sethu:
PR just adds a bunch of utility functions + a testing framework in order to prepare for the immense amount of integration tests we should probably write :).


backend/editor/OT/testing_framework.go
```diff
@@ -0,0 +1,87 @@
...
+// testing_framework is a simple framework that can be used for performing integration tests on the concurrent editor
+// it allows for the observation of client behavior and tracking document server behavior
+type TestingClient struct {
+	underlyingClient *clientView
+	operationPipe    pipe
+	terminationPipe  func()
+}
+
+func (tC TestingClient) HasTerminated() bool   { return len(tC.underlyingClient.sendTerminateSignal) > 0 }
+func (tC TestingClient) WasAcknowledged() bool { return len(tC.underlyingClient.sendOp) > 0 }
```
hdphuong, line 21: why is it `sendOp` and not `sendAcknowledgement` here?
also, would it always be false since [the condition in select-case](https://github.com/csesoc/website/blob/36cfebae51a6b16888af5bc152b5eb4e03b93e0b/backend/editor/OT/client_view.go#L39) i.e. `<-c.sendOp` takes the message out of the channel?

hdphuong (changes requested):
!LGTM (sorry xD)

backend/editor/OT/testing_framework.go
Varun-Sethu: Oooh yup thanks for the catch, yeah it should be `sendAcknowledgement` also `WasAcknowledged` will always be false until we get an acknowledgement and from then on-wards it will be true, the point of this framework is to sorta "emulate" that run function without actually running it

although we do probably need a way for the framework to acknowledge the ack so ill add that in 😎

hdphuong (approved):

PR #302, opened 2022-11-21
New sponsors page

winnteas:
- created new sponsors UI
- added mobile responsiveness

Screen Shot 2022-11-21 at 2 05 47 pm
Screen Shot 2022-11-21 at 2 06 11 pm


next/src/pages/Sponsors2.tsx
```diff
@@ -0,0 +1,219 @@
...
+import Navbar from "../components/navbar/Navbar";
+import { NavbarOpenHandler, NavbarType } from "../components/navbar/types";
+import styled from 'styled-components'
+import Link from 'next/link'
+import { device } from '../styles/device'
+
+
+import Otter from '../svgs/otter.png'
+
+const Text = styled.p`
```
lauraw0, line 13: all these styled components can be moved into a different file and then imported into this one!

lauraw0 (changes requested):
this sponsors component should be at the end of the homepage and not as a separate page. so it should replace what we have in the MiniSupport.tsx. besides these small issues it looks great!! :)

PR #303, opened 2022-11-21
Footer Icons

winnteas:
- added discord, facebook, instagram, youtube and spotify icons with links in the footer
Screen Shot 2022-11-21 at 2 09 00 pm
Screen Shot 2022-11-21 at 2 09 14 pm


next/src/components/footer/Footer.tsx
```diff
@@ -48,6 +67,24 @@ const Footer: React.FC<{}> = () => {
...
+            </a>
+            <a href="https://www.youtube.com/c/CSESocUNSW">
+              <Image src={YoutubeLogo} alt="Instagram" />
+            </a>
+            <a href="https://open.spotify.com/show/2h9OxTkeKNznIfNqMMYcxj">
+              <Image src={SpotifyLogo} alt="Instagram" />
+            </a>
+          </ImagesContainer>
           <br /><br />
           © 2021 — CSESoc UNSW
```
Varun-Sethu, line 89: should this be 2022?

next/src/components/footer/Footer.tsx
```diff
@@ -48,6 +67,24 @@ const Footer: React.FC<{}> = () => {
           B03 CSE Building K17, UNSW
           <br />
           csesoc@csesoc.org.au
+          <br/><br/>
+          <ImagesContainer>
+            <a href="https://discord.gg/AM4GB5zuB6">
```
lauraw0, line 72: pls change the <a> tags into next/link tags! so import the following `import Link from next/link` and simply just replace the `<a>` tags with `<Link>` tags :>

next/src/components/footer/Footer.tsx
lauraw0: aa yes it should be 2022! winnie could you also update the text into 2022 when making the requested changes :pray:

lauraw0 (changes requested):

next/src/components/footer/Footer.tsx
```diff
@@ -4,8 +4,27 @@ import Image from "next/image";
 import styled from "styled-components";
 
 import CSESocLogo from "../../../public/assets/logo_white.svg";
+import DiscordLogo from "../../../public/assets/socials/discord.svg";
+import FacebookLogo from "../../../public/assets/socials/facebook.svg";
+import InstagramLogo from "../../../public/assets/socials/instagram.svg";
+import YoutubeLogo from "../../../public/assets/socials/youtube.svg";
+import SpotifyLogo from "../../../public/assets/socials/spotify.svg";
```
zaxutic, line 11: thoughts on making these absolute imports? since you guys have [`"baseUrl": "."` in your tsconfig](https://github.com/csesoc/website/blob/main/frontend/tsconfig.json#L22), you should be able to just import from `"public/assets/logo_white.svg"` without needing to traverse all the way back up the tree (this will also help if you ever want to move this component or copy these imports somewhere else)

PR #304, opened 2022-11-22
docs: correct docker command from `volumes` to `volume`

zaxutic:
running the current command returns the following output:
```
docker: 'volumes' is not a docker command.
See 'docker --help'
```
because `volumes` is not a docker command, but `volume` is, so this PR fixes this docker command by changing `docker volumes` to `docker volume` so that it uses the correct docker command that doesn't return an error, as `docker volume` is valid


Varun-Sethu (approved):
🐰 🥕 !!!! 🌔 🚀 
(real)

PR #305, opened 2022-11-24
OT Integration - Part 2

Varun-Sethu:


zaxutic (commented):
so true frfr ong

frontend/src/packages/editor/operationManager.tsx
```diff
@@ -30,4 +30,98 @@ export const slateToCmsOperation = (editorContent: BlockData, operation: BaseOpe
             noop: {}
         }
     }
+}
+
+
```
zaxutic, line 35: why 2 blank lines? this isn't python/pep 8 😭

frontend/src/packages/editor/operationManager.tsx
```diff
@@ -30,4 +30,98 @@ export const slateToCmsOperation = (editorContent: BlockData, operation: BaseOpe
...
+// for example: if we receive the slate operation {set-node underline = true of text-object} with the path [0, 0, 0]
+// the normalised path will be [0, 0, 0, 4] (assuming 4 is index 4 in the text object)
+const normalizePath = (editorContent: BlockData, op: NodeOperation | TextOperation): number[] => {
+    // quirkiness with the CMS text operation interface, the target is included in the path
+    if (op.type === "remove_text" || op.type === "insert_text") {
+        return op.path;
+    }
+
+    // resolve what type we're studying and fetch the field mappings for it
+    return []
```
zaxutic, line 54:
```suggestion
    return [];
```

frontend/src/packages/editor/operationManager.tsx
```diff
@@ -30,4 +30,98 @@ export const slateToCmsOperation = (editorContent: BlockData, operation: BaseOpe
...
+        OperationType: "insert",
+        IsNoOp: false,
+        Operation: {
+            $type: "stringOperation",
+            stringOperation: {
+                rangeStart: op.offset,
+                rangeEnd: op.offset + op.text.length,
+                newValue: op.text
+            }
+        }
```
zaxutic, line 69:
```suggestion
                newValue: op.text,
            },
        },
```

frontend/src/packages/editor/operationManager.tsx
```diff
@@ -30,4 +30,98 @@ export const slateToCmsOperation = (editorContent: BlockData, operation: BaseOpe
...
+        OperationType: "delete",
+        IsNoOp: false,
+        Operation: {
+            $type: "stringOperation",
+            stringOperation: {
+                rangeStart: op.offset - op.text.length,
+                rangeEnd: op.offset,
+                newValue: op.text,
+            }
+        }
```
zaxutic, line 86:
```suggestion
            },
        },
```

frontend/src/packages/editor/operationManager.tsx
```diff
@@ -30,4 +30,98 @@ export const slateToCmsOperation = (editorContent: BlockData, operation: BaseOpe
...
+    }
+);
+
+
+const normalizeElementPath = (contentBlock: CMSEditorNode, op: NodeOperation | TextOperation): number[] => {
+    switch (true) {
+        case IsCustomElement(contentBlock): return normalizeCustomElementPath(contentBlock as CustomElement, op);
+        case IsCustomTextBlock(contentBlock): return normalizeCustomTextPath(contentBlock as CustomText, op);
+    }
```
zaxutic, line 95: interesting

i feel like if you want each branch on a single line here it's easier to just do something like
```suggestion
    if (IsCustomElement(contentBlock)) return normalizeCustomElementPath(contentBlock as CustomElement, op);
    if (IsCustomTextBlock(contentBlock)) return normalizeCustomTextPath(contentBlock as CustomText, op);
```

i think this is a go habit?
(note that we don't have to worry about else-if here because they all return anyway)

frontend/src/packages/editor/operationManager.tsx
```diff
@@ -30,4 +30,98 @@ export const slateToCmsOperation = (editorContent: BlockData, operation: BaseOpe
...
+        "link": 1,
+        "bold": 2,
+        "italic": 3,
+        "underline": 4
+    }
+
+    const isTextOp = op.type === "insert_text" || op.type === "remove_text";
+    const finalIndex = isTextOp
+                        ? fieldIndexes.text
+                        :
```
zaxutic, line 125: ❓

zaxutic (changes requested):

frontend/src/packages/editor/operationManager.tsx
zaxutic: SO TRUE

frontend/src/packages/editor/operationManager.tsx
Varun-Sethu: Michael is a draft PT to showcase an idea 😭

frontend/src/packages/editor/operationManager.tsx
```diff
@@ -30,4 +30,98 @@ export const slateToCmsOperation = (editorContent: BlockData, operation: BaseOpe
...
+
+    return [];
+}
+
+const normalizeCustomElementPath = (contentBlock: CustomElement, op: NodeOperation | TextOperation): number[] => {
+    return [] 
+}
+
+const normalizeCustomTextPath = (contentBlock: CustomText, op: NodeOperation | TextOperation): number[] => {
+    // TODO: For Mae :P - So the issue here is that fields on javascript objects are unordered, this is unlike Go where (typically) the order in which
```
zaxutic, line 105: this isn't quite correct, objects do maintain some order:

> The traversal order, as of modern ECMAScript specification, is well-defined and consistent across implementations. Within each component of the prototype chain, all non-negative integer keys (those that can be array indices) will be traversed first in ascending order by value, then other string keys in ascending chronological order of property creation.

([MDN for..in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#description), this also extends to object methods like `Object.keys()` and `Object.entries()`)

Varun-Sethu: Oh  thats convenient, so should it directly correspond to the the order declared in the TS type?

Varun-Sethu: > Oh thats convenient, so should it directly correspond to the the order declared in the TS type?

Wait ig not since it depends on order of declaration in the prototype so ig anyone can declare it in any order they want

zaxutic: not necessarily as declared in the type, as those aren't available at runtime - if you send an json object from the backend and parse that from the frontend, it *should* maintain whatever order it was sent in (assuming there's no integer keys)

Varun-Sethu: Hmm issue is though is that slate doesn't actually seem to give us a guarantee on order (for starters since it doesn't surface the entire object that its update and only surfaces the field)

zaxutic: gg rip that's doomed then

PR #306, opened 2022-12-24
Starter Guide - Enrolment Component

winnteas:
- implemented enrolment UI in the starter guide

image
image


next/yarn.lock
```diff
@@ -0,0 +1,2876 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
zaxutic, line 1: wait is this file meant to be here, there's already a package-lock.json (we use npm here?)

PR #335, opened 2023-03-04
[WEB-10] Handle enter key press when naming new file/folder

derek-ht:
When creating a new file/folder the popup modal does not accept the enter key press to submit. Handling the enter key press provides better accessibility.
### Changes
- added an onKeyDown event to the submission textfield in ConfirmationWindow.tsx
- created the event handler (handleKeyDown) that calls handleSubmit when pressed
### Comments


lauraw0 (approved):
lgtm you can merge now! :D

PR #337, opened 2023-03-06
Filled in the testEditHandler, but it will send back a 500 error beca…

Expressionless-Ball-Thing:
but it will send back a 500 error because of the websocket upgrade thing.

Not sure how to fix it.

Also I still have no idea how any of this websocket business is suppose to work.


hdphuong (approved):
LGTM!

PR #338, opened 2023-03-12
remove input box when user deselects the title of a folder or a file

winnteas:
Why the changes are required

- the input box persists and the user may end up with multiple input boxes for changing the names of the files and folders

Changes

- the input box is automatically focused once the name is double clicked
- clicking outside the box will set the name


lauraw0 (approved):
lgtm!

PR #340, opened 2023-03-13
OT Editor - Documentation

Varun-Sethu:
Adds documentation for the OT editor 😎


backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
+# The Concurrent Editor
+This is a rather full on package so this README document will outline how exactly the (OT) Concurrent Editor. As a high level overview though the concurrent editor is an OT based editor thats heavily influenced by Google's WAVE algorithm, Wave was Google's predecessor to Google docs and heavily inspired the architecture for Google Docs. Our editor uses the same underlying architecture as Wave except the difference lies in the type of operations we are sending back and forth between the server. Wave was originally designed to operate with operations that modified unstructured data, our operations modify structured JSON data.
```
angary, line 2:
```suggestion
As a high level overview though the concurrent editor is an OT based editor thats heavily influenced by Google's Wave algorithm, Wave was Google's predecessor to Google docs and heavily inspired the architecture for Google Docs. Our editor uses the same underlying architecture as Wave except the difference lies in the type of operations we are sending between the server and client. Wave was designed to operate with operations that modified unstructured data, while our operations modify structured JSON data.
```

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
...
+ - **The document server layer**
+    - The server layer is a big object that models a singular document being edited, it maintains a list of all active client objects that are connected to it and the current state of its AST.
+
+![Editor Architecture](./docs/editor_arch.png)
+
+## Starting a Document
+I personally feel like its rather easy to understand a system if you understand how it achieves its key features. This section is about what exactly happens when a user clicks the "edit" button on a document and constructs an edit session. What type of objects are created? Where do they live? Etc. 
+
+### The connection starts
+So the user clicks the "edit" button, this instantiates a HTTP request thats handled by the HTTP handler in `main.go`: `func EditEndpoint(w http.ResponseWriter, r *http.Request)`. This handler takes the incoming request, looks up the requested document and if it exists upgrades the connection to a Websocket connection.
```
angary, line 21:
```suggestion
When the user clicks the "edit" button, this instantiates a HTTP request that's handled by the HTTP handler in `main.go`: `func EditEndpoint(w http.ResponseWriter, r *http.Request)`. This handler takes the incoming request, looks up the requested document and if it exists upgrades the connection to a Websocket connection. 
```

hdphuong (approved):
lgtm if lgt Garry 😄

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
...
+	requestedDocument := // parse request body
+	targetServer := GetDocumentServerFactoryInstance().FetchDocumentServer(requestedDoc)
+
+	wsClient := newClient(ws)
+	commPipe, terminatePipe := targetServer.connectClient(wsClient)
+
+	go wsClient.run(commPipe, terminatePipe)
+}
+```
+Note that during the connection process the document server ended up returning a `pipe`, this `pipe` is actually just a small function that the client (or at least the client's shadow on the server) can use to propagate messages to the server it is connected to.
```
angary, line 37:
```suggestion
During the connection process the document server returned two "pipes". A "pipe" is a function that the client (or at least the client's shadow on the server) can use to propagate messages to the server it is connected to. 
```

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
...
+				signalDeparturePipe()
+				c.socket.Close()
+			}
+		}
+	}
+}
+```
+The bit of interest is what happens in the `default` branch of the select statement (we will talk about the other branches later). Within this branch we attempt to read something from the websocket and then parse that (we will cover parsing a little later as its surprisingly complicated), we then use the `serverPipe` mentioned previously to send that request to the document server. 
+
+This `serverPipe` is a function built by the `document_server` during connection it is a closure returned by the `buildClientPipe` method. The function is relatively intense and hands on so a lot of details have been left out here
```
angary, line 77:
```suggestion
This `serverPipe` is a closure returned by the `buildClientPipe` method during the connection setup with the `DocumentServer`. The function is relatively intense so a lot of details have been left out here.
```

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
...
+			s.clientsLock.Unlock()
+		}
+	}
+}
+```
+so whenever we get an operation from the client we:
+ 1. communicate it to the document_server via a pipe
+ 2. the operation is then transformed against the entire log of operations the server has applied
+ 3. the operation is then applied to the server's representation of the document
+ 4. the operation is then communicate to all other clients
```
angary, line 137:
```suggestion
So whenever we get an operation from the client we:
 1. communicate it to the document_server via a pipe
 2. the operation is then transformed against the entire log of operations the server has applied
 3. the operation is then applied to the server's representation of the document
 4. the operation is then communicated to all other clients 
```

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
...
+```go
+type clientView struct {
+	socket *websocket.Conn
+
+	sendOp              chan operations.Operation
+	sendAcknowledgement chan empty
+	sendTerminateSignal chan empty
+}
+```
+`sendOp`, the `wsClient.run` function is constantly listening for messages down these channels and actions on them accordingly.
```
angary, line 150: Maybe something was deleted here? Not too sure what it's trying to say

backend/editor/OT/docs/operation_parsing.md
```diff
@@ -0,0 +1,103 @@
+# Operation Parsing & Application
+
+Surprisingly most of the complexity in the OT editor lives in the operation parsing and application logic. This document aims to help demystify some of it. As a quick aside, everything regarding operations can be found in the `operation` folder.
+
+## The operation struct
+Operations (in Go) are defined as a struct, the operations we receive from the client conform to this structure and are parsed using a JSON parsing library (more on that later as there is a bit of complexity behind this).
```
angary, line 6:
```suggestion
Operations (in Go) are defined as a struct. The operations we receive from the client conform to this structure and are parsed using a JSON parsing library (more on that later as there is a bit of complexity behind this). 
```

backend/editor/OT/docs/operation_parsing.md
```diff
@@ -0,0 +1,103 @@
...
+
+Surprisingly most of the complexity in the OT editor lives in the operation parsing and application logic. This document aims to help demystify some of it. As a quick aside, everything regarding operations can be found in the `operation` folder.
+
+## The operation struct
+Operations (in Go) are defined as a struct, the operations we receive from the client conform to this structure and are parsed using a JSON parsing library (more on that later as there is a bit of complexity behind this). 
+```go
+type (
+	// EditType is the underlying type of an edit
+	EditType int
+
```
angary, line 11:
```suggestion
```

backend/editor/OT/docs/operation_parsing.md
```diff
@@ -0,0 +1,103 @@
...
+	// Operation is the fundamental incoming type from the frontend
+	Operation struct {
+		Path                  []int
+		OperationType         EditType
+		AcknowledgedServerOps int
+
+		IsNoOp    bool
+		Operation OperationModel
+	}
+)
```
angary, line 27:
```suggestion
)
// EditType is an enum with `int` as the base type
type EditType int
const (
	Add EditType = iota
	Remove
)
```
Would be good to show what the values of `EditType` are :)

backend/editor/OT/docs/operation_parsing.md
```diff
@@ -0,0 +1,103 @@
...
+```
+The above code snippet uniquely defines an operation. Operations take a path to where in the document AST they are being applied (see the paper on tree based transform functions) and the physical operation being applied. The operation being applied (`OperationModel`) is actually an interface and is the reason why parsing operations is more complex than it seems. This interface defines two functions, one for transforming a operation against another operation and one for applying an operation to an AST. 
+
+The reason why `OperationModel` is an interface is because there are several distinct types of operations we can apply for varying types. Theres different operations for editing an `integer`, `array`, `boolean`, `object` and `string` field. Each of these define their own transformation functions and application logic so in order to maintain a clean abstraction we use an interface. As an example this is how the `string` operation type implements this interface ![here](../operations/string_operation.go), its a rather intense implementation since it also implements string based transform functions.
```
angary, line 31:
```suggestion
The reason why `OperationModel` is an interface is because there are several distinct types of operations we can apply for varying types. Theres different operations for editing an `integer`, `array`, `boolean`, `object` and `string` field. Each of these define their own transformation functions and application logic, so in order to maintain a clean abstraction we use an interface. As an example this is how the `string` operation type implements this interface ![here](../operations/string_operation.go), its a rather intense implementation since it also implements string based transform functions.
```

backend/editor/OT/docs/operation_parsing.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+## Operation application
+Recall that the `document_server` maintains an abstract syntax tree for the current JSON document, this AST is exactly what `cmsjson.AstNode` is, the document server maintains the **root node**. When applying an operation to a document we invoke the `Operation.ApplyTo` function
```
angary, line 35:
```suggestion
Recall that the `document_server` maintains an abstract syntax tree for the current JSON document, this AST is implemented by `cmsjson.AstNode`, and the document server maintains the **root node**. When applying an operation to a document we invoke the `Operation.ApplyTo` function
```

angary (approved):
LGTM - just one line that seemed to be missing some words :)

Varun-Sethu (commented):
review change

hanyuone:
lgtm, main thing is i noticed in a lot of places the docs say "it's complicated" or "it's complex" - these are fluff sentences that imo not only don't add to documentation/explanation but make it harder for people reading to engage. will provide more detailed review as well

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
+# The Concurrent Editor
+As a high level overview though the concurrent editor is an OT based editor thats heavily influenced by Google's Wave algorithm, Wave was Google's predecessor to Google docs and heavily inspired the architecture for Google Docs. Our editor uses the same underlying architecture as Wave except the difference lies in the type of operations we are sending between the server and client. Wave was designed to operate with operations that modified unstructured data, while our operations modify structured JSON data.
```
hanyuone, line 2: "As a high level overview ~~though~~, the concurrent editor"

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+Before reading the rest of this document I highly recommend you read through the resources linked in the base [README.md](../../README.md).
+
+## High Level Architecture
+At a very high level our editor server consists of 3 distinct layers.
+ - **The data layer**
+    - This layer is the server's copy of the current state of the document. This layer is what all incoming operations will modify. Our initial design for the data layer involved a singular struct modelling the entire document that we modified using reflection. This proved tricky due to the intricacies of Go's reflection system so we moved to an AST based approach. Currently the data layer is just the AST for the JSON of the document, and operations modify this AST directly. To prevent corrupted documents we have various data integrity checks utilising reflection.
```
hanyuone, line 9: is it necessary to mention reflection (i.e. is it still in the codebase)? if you really want to include it, it could be in some other document/section, but not here if the old approach is completely removed

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
...
+ - **The document server layer**
+    - The server layer is a big object that models a singular document being edited, it maintains a list of all active client objects that are connected to it and the current state of its AST.
+
+![Editor Architecture](./docs/editor_arch.png)
+
+## Starting a Document
+I personally feel like its rather easy to understand a system if you understand how it achieves its key features. This section is about what exactly happens when a user clicks the "edit" button on a document and constructs an edit session. What type of objects are created? Where do they live? Etc. 
+
+### The connection starts
+When the user clicks the "edit" button, this instantiates a HTTP request that's handled by the HTTP handler in `main.go`: `func EditEndpoint(w http.ResponseWriter, r *http.Request)`. This handler takes the incoming request, looks up the requested document and if it exists upgrades the connection to a WebSocket connection. This is important as a WebSocket connection allows for bidirectional communication between the client and server  in real time without needing either to poll for updates
```
hanyuone, line 21: full stop at the end

backend/editor/OT/docs/operation_parsing.md
```diff
@@ -0,0 +1,106 @@
...
+	}
+)
+// EditType is an enum with `int` as the base type
+type EditType int
+const (
+	Add EditType = iota
+	Remove
+)
+```
+The above code snippet uniquely defines an operation. Operations take a path to where in the document AST they are being applied (see the paper on tree based transform functions) and the physical operation being applied. The operation being applied (`OperationModel`) is actually an interface and is the reason why parsing operations is more complex than it seems. This interface defines two functions, one for transforming a operation against another operation and one for applying an operation to an AST.
```
hanyuone, line 32: would be nice to link the papers in the docs themselves, saves anyone reading having to click back and forth

backend/editor/OT/docs/operation_parsing.md
```diff
@@ -0,0 +1,106 @@
...
+type EditType int
+const (
+	Add EditType = iota
+	Remove
+)
+```
+The above code snippet uniquely defines an operation. Operations take a path to where in the document AST they are being applied (see the paper on tree based transform functions) and the physical operation being applied. The operation being applied (`OperationModel`) is actually an interface and is the reason why parsing operations is more complex than it seems. This interface defines two functions, one for transforming a operation against another operation and one for applying an operation to an AST. 
+
+The reason why `OperationModel` is an interface is because there are several distinct types of operations we can apply for varying types. Theres different operations for editing an `integer`, `array`, `boolean`, `object` and `string` field. Each of these define their own transformation functions and application logic, so in order to maintain a clean abstraction we use an interface. As an example this is how the `string` operation type implements this interface ![here](../operations/string_operation.go), its a rather intense implementation since it also implements string based transform functions.
```
hanyuone, line 34: can get rid of the "it's a rather intense implementation" phrase

backend/editor/OT/docs/operation_parsing.md
```diff
@@ -0,0 +1,106 @@
...
+			"booleanOperation": reflect.TypeOf(BooleanOperation{}),
+			"stringOperation":  reflect.TypeOf(StringOperation{}),
+
+			"arrayOperation":  reflect.TypeOf(ArrayOperation{}),
+			"objectOperation": reflect.TypeOf(ObjectOperation{}),
+		},
+	},
+}
+```
+the configuration is in essence a mapping between the the `reflect.Type` representation of the interface and the `reflect.type` representation of every struct that "implements it". Implements is in quote as there is no way to statically verify this at compile time, instead if any of these config options are invalid a runtime error will be thrown during parsing. Usage of the `cmsjson` library for the most part is rather simple (thanks to generics in Go :D). An example can be found in the `ParseOperation` function within `operation_model.go`.
```
hanyuone, line 73: can include a link to the code with the example

backend/editor/OT/README.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+So whenever we get an operation from the client we:
+ 1. communicate it to the document_server via a pipe
+ 2. the operation is then transformed against the entire log of operations the server has applied
+ 3. the operation is then applied to the server's representation of the document
+ 4. the operation is then communicated to all other clients 
+
+### The document server wants to propagate operations
+If you remember previously how it was mentioned that the document server "propagates" operations to the clients? It does that by sending these operations down a channel maintained by each client
```
hanyuone, line 140: rewrite sentence, doesn't make sense

hanyuone (changes requested):

PR #342, opened 2023-03-19
Fix test enivronment determination

angary:
This change no longer requires the `-v` flag to be present when testing as it simply checks whether the flag has been added as an environment variable during the loading of the testing package, rather than if it has been set.

Seems to be the suggested way of checking if we are in a testing environment according to [stack overflow](https://stackoverflow.com/questions/14249217/how-do-i-know-im-running-within-go-test).


Varun-Sethu:
Do you mind removing the `-v` flag from the `go test` in our github actions pipeline? https://github.com/csesoc/website/blob/main/.github/workflows/ci.yml#L28 doing so will allow you to verify if the changes actually work

Varun-Sethu (approved):
LGTM (the looks good part, not the lets get this merged part)

lhjt (approved):

hdphuong (approved):

ShunyaoLiang (approved):

PR #343, opened 2023-03-25
[WEB-20] Clicking on folder pill in breadcrumb path redirects user to that folder

brian2w:
Breadcrumb pill path was not interactive and did not allow user to traverse the folder structure by clicking on it.
Repeatedly clicking on the root folder also caused the path to repeatedly print out 'root'.

### Changes
- Implemented breadcrumb pill clicking functionality by altering how path was stored: string path -> array of objects storing folderId and folderNames
- Using the above change to an array, the repeating root in path was also fixed by checking if a certain folderId already existed in the path array.
### Comments
Now that the path is stored in an array of objects, the traverseBackFolderSaga can simply be replaced by traverseIntoFolderSaga, so can potentially be removed.


lauraw0 (approved):
lgtm! merge whenever :)

PR #347, opened 2023-04-06
Web27 Display Filename

winnteas:
Why the changes are required

- to see what file is being edited on the editor page

Changes

- passed in filename when navigating to the editor page (via edit button or double clicking)
- filename shows on the left side of the header, buttons remain on the right side

image


james-teng-0 (dismissed):
lgtm but please fix the frontend_testing pipeline before merging  - Error: Uncaught [Error: useLocation() may be used only in the context of a Router component.]

- you probably need to wrap the parent element in the Router component.

james-teng-0:
I think you can use a parent useState in Dashboard.tsx and pass it to the Renderer and Sidebar so that you can pass the name to the Edit button. 
Similar to how the selectedFile / setSelectedFile is implemented.

frontend/src/deprecated/components/Editor/EditorHeader.tsx
```diff
@@ -4,6 +4,7 @@ import Button from '@mui/material/Button';
 import IconButton from "@mui/material/IconButton";
 import ArrowBackIcon from "@mui/icons-material/ArrowBack";
 import { useNavigate } from 'react-router-dom';
+import { ReactNode } from 'react'
```
lauraw0, line 7: pls remove unused imports!

frontend/src/packages/dashboard/Dashboard.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from "react";
-import { useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
```
lauraw0, line 2: useSelector is not used here so it can be removed :>

lauraw0 (approved):
just some style changes needed but besides that lgtm! you can merge once you have removed those unused imports

frontend/src/packages/dashboard/components/SideBar/SideBar.tsx
```diff
@@ -113,6 +116,12 @@ export default function SideBar({
   isOpen,
   setOpen,
 }: Props) {
+
+    const folders = useSelector((state: RootState) =>
+      folderSelectors.getFolderState(state)
+    ).items;
+
+
```
lauraw0, line 124: You can replace this code with the existing getFolderState() function in `/frontend/src/packages/dashboard/api/helpers.ts` !

PR #353, opened 2023-05-11
[WEB-26] Create delete function for removing files and folders

brian2w:
Currently, there is no way to delete files and folders in the CMS.

### Changes
- Added delete file button to sidebar
- Added delete confirmation modal
- Opening a folder is now a double click (rather than a single click)
- Single click 'selects' the folder (indicator still needed)

### Screenshots
image

### Comments
- Need to contact backend team regarding recursively deleting contents inside a folder

### DIFF
--- a/frontend/src/packages/dashboard/Dashboard.tsx
+++ b/frontend/src/packages/dashboard/Dashboard.tsx
@@ -1,19 +1,19 @@
-import React, { useEffect, useState } from "react";
-import { useDispatch } from "react-redux";
-import styled from "styled-components";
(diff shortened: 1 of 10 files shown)


lauraw0 (changes requested):
Looks great!! :D We'll ignore the changes in whitespace for now (creating a prettier template is on my todo list somewhere) but if you could make these small ux changes that would be awesomeee:

- when the user clicks away from the folders and files (onBlur) the selectedFile should revert to null
- when the user presses on the delete button on the keyboard (onKeyPress) the same delete functionality should apply

PR #355, opened 2023-05-24
[WEB-67] [CMS] rename redux function

brian2w:
Why the changes are required?
Currently, there is no api call to the backend to rename fileEntity names.

Changes
Added rename saga which calls backend api.

Screenshots

Comments


lauraw0 (approved):
LGTM! merge whenever~~

PR #356, opened 2023-05-25
Context Refactoring

brianzhang2101:
Moving the construction of contexts outside of `Get___Repo()` functions into a parameter instead. This fixes a use case where uninitialised `Repo`s have a cyclical dependency on `TestingContext` (It's required to be defined, but cannot be at the same time).


Varun-Sethu (approved):
Looks good to me! Make sure some of the endpoints still work though

hdphuong (approved):

PR #374, opened 2023-06-17
[WEB-65] [EDITOR] Make blog title editable within editor

josh-ramos-22:
## Why the changes are needed
Currently there is no way to edit a document's filename while inside the editor.

## Changes
I took inspiration from Google Docs and made it so that when a user hovers over the filename in the editor, the cursor turns into an I-beam and the title slightly enlarges. This indicates to the user that the title is editable. Then upon clicking, the title visibly becomes a text box which users can write into. Once they are done renaming, they click out of the text box and the new filename is saved :D 

### File changes summary
- `frontend/src/cse-ui-kit/EditableTitle_textbox/*` - new component defining the custom text box for the editable file name
- `frontend/src/packages/editor/index.tsx` - integrated code to support file renaming. Reused code from `frontend/src/packages/dashboard/components/FileRenderer/Renamable.tsx` to update the redux state to ensure the new file name is propagated globally.

### Demo

https://github.com/csesoc/website/assets/79197816/4ce878c9-88cd-4abb-9dcf-664f2514bac5

### DIFF
--- a/frontend/src/cse-ui-kit/EditableTitle_textbox/EditableTitle-Styled.tsx
+++ b/frontend/src/cse-ui-kit/EditableTitle_textbox/EditableTitle-Styled.tsx
@@ -1,4 +1,5 @@
 import styled from "styled-components";
+import { InputHTMLAttributes } from "react"
 
 export type buttonProps = {
   background?: string;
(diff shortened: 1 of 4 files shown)


lauraw0 (approved):
lookss niceeee

PR #376, opened 2023-07-11
[WEB-66] Fix Files and Folders Don't Wrap When Resizing Screen

brian2w:
Currently the files and folders overflow off the side of the screen (as shown below). They should instead wrap around in a sort of grid manner.

### Changes
- Folders and Files now wrap around when screen is resized
- Folders and files line up in a grid system, where each takes 140px by 161px
### Screenshots
image

### Comments


lauraw0 (approved):
beautiful

PR #377, opened 2023-07-13
[WEB-31] Bug renaming reordering issue (and also WEB-68)

josh-ramos-22:
## Why are the changes needed?
Currently, when you rename a file in the folder and refresh, the renamed file is repositioned to the end of the file list, jumbling the original order.

### WEB-68
In this ticket I also happened to fixed the issue covered in the ticket WEB-68 (https://csesoc.atlassian.net/browse/WEB-68) where input boxes were rendered with old initial values.

## Changes
The reason that files are reordered upon renaming is due to how `UPDATE`s in PostgreSQL works. PostgreSQL tables store their records in an undefined order, hence we cannot assume that a table retains its original order after an `UPDATE`.

A possible fix to this would to be adding an explicit ordering to the query that fetches the files from the database. However this seems infeasible as there does not seem to be an existing `datetime_created` or `serial` field in the existing database schema to explicitly sort the files on the desired order.

This PR contains an alternate solution which displays files in alphabetical order. Additionally, the comparator used also ensures that all folders are displayed before files.

### WEB-68
The reason why initial input values were not updating was because these elements are actually rendered multiple times upon refresh.

First, the file containers render with the data stored on the frontend in redux, which is in the initial order before the renaming of the file.  Then, after the call to the backend, these values are updated with the new order. However, since the files will be in a new order and  initial `useState()` values are only considered during the first render, the initial values will not be updated. This causes a mismatch between file names and initial values.

To fix this, I explicitly called `setInputName(name)` in the event handler for when the file name label is double clicked to ensure that the proper initial value is displayed.

### DIFF
--- a/frontend/src/packages/dashboard/api/index.ts
+++ b/frontend/src/packages/dashboard/api/index.ts
@@ -31,7 +31,6 @@ export async function updateContents(id: string) {
   const children_json = await children_resp.json();
   const children = children_json.Response.Children.map(
     (child: JSONFileFormat) => {
-      console.log(child);
       return toFileOrFolder(child);
     }
   );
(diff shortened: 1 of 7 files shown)


lauraw0 (approved):
lookss amazinggg merge whenever

PR #382, opened 2023-08-26
[WEB-25] Implement checklists

winnteas:
**Why the changes are required?**

Users may want to apply a checklist within the editor similar to the numbered and bulleted lists

**Changes**

- Added checklist button next to quote button
- Created checklist element tag in EditorBlock.tsx
image


frontend/src/packages/editor/components/EditorBlock.tsx
```diff
@@ -46,6 +48,51 @@ const Text = styled.span<{
...
+          style={{
+            marginRight: '0.75em'
+          }}
+        >
+          <input
+            type="checkbox"
+            checked={checked}
+            onChange={event => {
+              // Figure out why ReactEditor.findPath does not work
+              // const path = ReactEditor.findPath(editor as ReactEditor, leaf);
```
lauraw0, line 72: Your current implementation works so you can just remove these comments but if you wanted to know why the ReactEditor.findPath() doesn't work, i think these could possibly be the reasons:
1. the editor needs to be passed in from the EditorBlock component
2. this findPath function is specifically finding a **node** (https://docs.slatejs.org/libraries/slate-react/react-editor#reacteditor.findpath-editor-reacteditor-node-node-path) but here you are rendering the checklists as leaves. This means that the function will never return the current **leaf**. You can resolve this by changing the checklists into nodes. NOTE: you can think of a node as sort of a big wrapper/container whilst leaves are sort of the children with decorations. However, I do have some doubts about this cause in that case it should in theory find the leaf level text node surrounding this lead? Not completely sure without delving into it too deeply.
3. This is probably the most likely reason: the structure of our editor/leaves/nodes does not follow typical slate operations and therefore any functions that involve traversing through the editor path is doomed to fail

frontend/src/packages/editor/components/EditorBlock.tsx
```diff
@@ -46,6 +48,51 @@ const Text = styled.span<{
...
+
+const CheckListItemElement = ({children, attributes, leaf}: any) => {
+  const readOnly = useReadOnly()
+  const [checked, setChecked] = React.useState(false)
+  return (
+    <div {...attributes} style={{display: 'flex', alignItems: 'left'}}>
+      <div
+        style={{ userSelect: "none" }}
+        contentEditable={false}
+      >
```
lauraw0, line 60: This div doesn't appear to do much so you can remove it

frontend/src/packages/editor/components/EditorBlock.tsx
```diff
@@ -73,9 +120,17 @@ const EditorBlock: FC<CMSBlockProps> = ({
         code: leaf.code ?? false,
         align: leaf.align ?? 'left',
         textSize: leaf.textSize ?? defaultTextSize,
+        checklist: leaf.checklist ?? false,
+        checked: leaf.checklist ?? false,
```
lauraw0, line 124: This doesn't appear to be used elsewhere so you can remove it. Let me know if i missed its use case

lauraw0 (changes requested):
nice nice looks pretty dang good!! just a couple changes needed (as detailed in the comments) plus the following notes:

1. Pls change it so that the checkboxes are reflected in the published blogs. This may possibly require some refactoring as you will need to not only keep track of the type (e.g. checklist, quote, code....) but also whether the items are checked/crossed out. This means you may need to add an optional type to the type interfaces.
**Editor**
image
**Published blog**
image
2. Sort of tied to the above point. After you have corrected the interface you will need to ensure that checked items remains checked after refresh. Currently your checked items become unchecked due to the checked value being held in a usestate.

PR #383, opened 2023-09-08
Fix parsing form file reflection error

angary:
Originally there as an error parsing uploaded image requests. This is because `field.Type().Name()` returns a `string` of the field type and so an error was thrown from trying to read a non-existing file from the request.

Correcting this to `v.Type().Field(i).Name` changes it to a `string` of the field name.


ShunyaoLiang (approved):
Straightforward. Love it

